PR #14, opened 2016-12-14
Add Json.toBytes to convert a JsValue directly to a byte array

gmethvin:
See also playframework/playframework#6334

### DIFF
--- a/play-json/src/main/scala/play/api/libs/json/Json.scala
+++ b/play-json/src/main/scala/play/api/libs/json/Json.scala
@@ -60,6 +60,14 @@ object Json {
    */
   def stringify(json: JsValue): String = JacksonJson.generateFromJsValue(json)
 
+  /**
+   * Convert a JsValue directly to an array of bytes (using UTF-8 encoding)
+   *
+   * @param json the JsValue to convert
+   * @return an Array[Byte] with the UTF-8-encoded JSON representation
+   */
+  def toBytes(json: JsValue): Array[Byte] = JacksonJson.jsValueToBytes(json)
+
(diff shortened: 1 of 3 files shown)


marcospereira:
Should this be also added to the Java API?

gmethvin:
@marcospereira Actually, we might not even want to keep `play.libs.Json` in the `play-json` project, since it's only a thin wrapper for Jackson. I just moved it because it was in the play-json module already, but it probably makes more sense to have it in `playframework` since it's not really a library on its own and has nothing to do with the Scala JSON library.

marcospereira:
Ok, will you add the `Writeable` change here?

gmethvin:
`Writeable` is a type class used in Play Scala actions to automatically convert types to an `HttpEntity` to be sent as part of a `Result`. It doesn't make sense to include here since that's the responsibility of the main Play library, not play-json.

play-json/src/main/scala/play/api/libs/json/Json.scala
```diff
@@ -60,6 +60,14 @@ object Json {
    */
   def stringify(json: JsValue): String = JacksonJson.generateFromJsValue(json)
 
+  /**
+    * Convert a JsValue directly to an array of bytes (using UTF-8 encoding)
```
cchantep: Indent issue? (https://github.com/playframework/play-json/pull/18/commits/aa1776a37d1ad56be584fb26b7b94e4f1c1e351f)

gmethvin:
@cchantep I think this is good now. Afterward we could release M2 and update #6334.

cchantep:
Would be nice to also have #13 in a next release.

PR #16, opened 2016-12-18
Support sealed trait with JSON macros

cchantep:


play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -256,227 +363,273 @@ object JsMacroImpl {
...
-        Implicit(name, ptype, neededImplicit, tpe, selfRef)
+          case _ => $json.JsError("error.expected.jsobject")
+        }"""
       }
 
-      // To print the implicit types in the compiler messages
-      private def prettyType(t: Type): String =
-        boundTypes.getOrElse(t.typeSymbol.fullName, t) match {
-          case TypeRef(_, base, args) if args.nonEmpty => s"""${base.asType.fullName}[${args.map(prettyType(_)).mkString(", ")}]"""
+      def writeLambda: Tree = {
```
cchantep: The `OWrites` is tricky to generate, due its contravariance

play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -256,227 +363,273 @@ object JsMacroImpl {
...
-        val neededImplicitType = appliedType(natag.tpe.typeConstructor, ptpe)
-        val neededImplicit = if (!selfRef) {
-          c.inferImplicitValue(neededImplicitType)
-        } else c.untypecheck(
-          // Reset the type attributes on the refactored tree for the implicit
-          ImplicitTransformer.transform(
-            c.inferImplicitValue(neededImplicitType))
-        )
+        q"""(_: $json.JsValue) match {
+          case obj @ $json.JsObject(_) => obj.value.get("_type") match {
```
gmethvin: Maybe this field should be configurable in some way and not hard-coded?

cchantep: Yes, but not sure it should be done right in this PR, as it may be related with some refactoring of `JsonConfiguration`

cchantep:
Review welcome

play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -263,236 +371,286 @@ object JsMacroImpl {
...
-        val resolvedImplicits = params.map { param =>
-          createImplicit(param.name, param.typeSignature)
-        }
-        val effectiveImplicits = if (ApplyUnapply.hasVarArgs) {
-          val varArgsImplicit = createImplicit(
-            resolvedImplicits.last.paramName,
-            ApplyUnapply.unapplyReturnTypes.get.last
-          )
+        q"""(_: $json.JsValue) match {
+          case obj @ $json.JsObject(_) => obj.value.get("_type") match {
```
gmethvin: I think this could be pulled from `JsonConfiguration` by adding a `typeProperty: String` (or similar) in there and using `$cfgName.typeProperty`.

Using the class name could be problematic. Why should the JSON serialization of my class change if I move the class to another package? A good compromise might be to use the simple name of the class (`Foo` instead of `com.example.Foo`). You could also have a `classNaming` in `JsonConfiguration` that maps the `Class[_]` to a name.

Also we don't have to keep `JsonConfiguration` in it's current form since it's a new feature in 2.6.0. It would be great to hear if you have other ideas for handling configuration.

cchantep: Using `JsonConfiguration` is planned to customize the naming (name of the discriminator field, value for this field accord each case/type).

I personnally feel it could be done in a separate PR, to validate/merge step by step, as long as each step is validated and backward compatible.

cchantep:
@gmethvin About `Writes` writing the discriminator in all case, I agree it would be more compatible with the contravariance.

I'm not sure developer expect such behaviour, even if it wouldn't not break the `Read`.

That's out of scope, but considering the Scala macro/implicit issue with contravariance, I tend to think it doesn't deserve to be keep like this (except through the contrafunctor, `Encoder` in circe).

gmethvin:
@cchantep Regardless of contravariance, I think it makes sense for the discriminator property to be embedded in the object, at least as the default behavior. As a developer it seems natural that every property in the Scala object would correspond to the same property on the JSON object. It's also the most commonly-used strategy in my experience.

For the other stuff I'm okay with you following up with a different pull request.

cchantep:
If writing a type discriminator property is the default behaviour for the `Writes`, for all the cases, then I think a configuration option should be provided to disabled it, as it's useless the majority of the case (100% of the existing use cases of Play JSON actually).

gmethvin:
@cchantep I'm not sure why you say it's useless in the majority of cases. I have seen an embedded type property of some sort used in many real apps and APIs that use JSON, much more so than the wrapper object strategy.

cchantep:
Personnally I have never seen such use case outside of sealed family, and I don't remember issue raise by user/developer about that.

gmethvin:
Well, most examples of Jackson polymorphic deserialization seem to use `JsonTypeInfo.As.PROPERTY`, and that is also [the default](https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization#12-per-class-annotations). It's also one of the suggested styles in [play-json-derived-codecs](https://github.com/julienrf/play-json-derived-codecs#custom-representation-of-sum-types). 

Not opposed to having multiple ways to do it with configuration though.

cchantep:
For these use cases correspond to when it's wanted to support a sealed family.

There is no technical issue in added a discriminated value for any type, even when it's asked to generate a `OWrites` directly. Personally I doubt it's neither useful outside of sealed family nor what's excepted.

cchantep:
**Default behaviour in other JSON libs**

[Fully automatic derivation](https://circe.github.io/circe/codec.html#fully-automatic-derivation) with Circe

[Lift JSON](https://github.com/lift/lift/blob/master/framework/lift-base/lift-json/README.md) (discriminator/`jsonClass` only for polymorphic serializer)

[Spray JSON](https://github.com/spray/spray-json/blob/master/src/test/scala/spray/json/ProductFormatsSpec.scala#L201)

gmethvin:
Hi Cedric. Just to make this clear, we're only talking about the sealed trait case, and I'm suggesting the `_type` property (which could eventually be configurable) only be added to the JSON representation for sealed traits.

spray-json-shapless also seems to prefer the "type" discriminator property: https://github.com/fommil/spray-json-shapeless/blob/master/src/test/scala/fommil/sjs/FamilyFormatsSpec.scala#L326

cchantep:
I have preference for the default naming of the discriminator field in case of sealed family, except `type` that looks confusing to me (even if Scala types should not have a `type` property).

gmethvin:
Yeah I think `_type` is fine as the default. Also makes it more readable since the _ communicates that it's not a public property of the object.

play-json/src/test/scala/play/api/libs/json/MacroSpec.scala
```diff
@@ -77,6 +120,34 @@ class MacroSpec extends org.specs2.mutable.Specification {
...
+        "_type" -> JsString("play.api.libs.json.MacroSpec.Optional")
+      )
+
+      lazy val wsimple = Json.toJson[Family](simple)
+      lazy val wopt = Json.toJson[Family](optional)
+
+      wsimple must_== jsSimple and {
+        wsimple.validate(Json.reads[Simple]).get must beTypedEqualTo(simple)
+      } and (wopt must_== jsOptional) and {
+        wopt.validate(Json.reads[Optional]).get must beTypedEqualTo(optional)
```
cchantep: @gmethvin Updated to have the `_type` discriminator at the same level as the other properties

play-json/src/test/scala/play/api/libs/json/MacroSpec.scala
```diff
@@ -63,11 +63,54 @@ class MacroSpec extends org.specs2.mutable.Specification {
...
+          "_type" -> "play.api.libs.json.MacroSpec.Simple",
+          "_value" -> Json.writes[Simple].writes(simple)
+        )
+
+        val jsOptional = Json.obj(
+          "_type" -> "play.api.libs.json.MacroSpec.Optional",
+          "_value" -> Json.writes[Optional].writes(optional)
+        )
+
+        jsSimple.validate[Family].get must beTypedEqualTo(simple) and {
```
gmethvin: So it seems like this is able to read both the verbose and compact syntax by default?

cchantep: Yes, as the `Reads` should anyway check for `_value` in case of non-`JsObject`, there is no extra cost to do it also for `JsObject`

Issue #10, opened 2016-12-12
Support sealed trait with JSON macros


gmethvin:
**(Moved from playframework/playframework#6424)**

Issue #4, opened 2016-12-12
Make Json.reads macro handle case class default params gracefully


gmethvin:
**(Moved from playframework/playframework#3244)**

It would be nice if `Json.reads` was able to synthesize `Reads` classes for `case class`s with default parameter values. So riffing on the example, we might have:

``` scala
case class Person(name: String, age: Int, lovesChocolate: Boolean, favoriteColor: String = "blue")

implicit val personReads = (
  (__ \ 'name).read[String] and
  (__ \ 'age).read[Int] and
  (__ \ 'lovesChocolate).read[Boolean] and
  ((__ \ 'favoriteColor).read[String] orElse Reads.pure("blue"))
)(Person)
```

(I imagine, there's probably some non-trivial reflection needed to get the default parameter value.)

Perhaps of interest to @mandubian

andyglow:
Hello, guys!
I have added a possible solution and I would like to have some feedback and wondering if it might be merged. Thanks!

fancywriter:
Hi @gmethvin!
Is it already in 2.6.0, as far as I understand?

fancywriter:
Hi @gmethvin , @andyglow 
I have tested it and actually it doesn't give me behaviour I expected.
```
import play.api.libs.json._

case class Foo(s: String = "default value")
implicit val format": Format[Foo] = Json.using[Json.WithDefaultValues].format[Foo]

println(Json.obj().validate[Foo])
// JsSuccess(Foo(default value),) <- OK
println(Json.obj("s" -> "overridden value").validate[Foo])
// JsSuccess(Foo(overriden value),) <- OK 
println(Json.obj("s" -> 100).validate[Foo])
// JsSuccess(Foo(default value),) <- NOT OK
println(Json.obj("s" -> Json.obj("foo" -> Json.arr("bar"))).validate[Foo])
// JsSuccess(Foo(default value),) <- NOT OK at all
```

It gives default value for any type of error - however, I expect to see it only for missing value or for null (it field is not Option of course, in this case I expect to see None for JsNull and error for everything else).

```
case class Foo(s: Option[String] = Some("default some"))

println(Json.obj().validate[Foo])
// JsSuccess(Foo(Some(default some)),)
println(Json.obj("s" -> JsNull).validate[Foo])
// works as expected JsSuccess(Foo(None),)
println(Json.obj("s" -> 100).validate[Foo])
[...]

PR #17, opened 2016-12-20
Add support for default values

andyglow:
Hello!
In our company we widely use play-json for years and it is great and convinient tool.
But recently we've came across the need in ability to restore case class state from partial json by applying default values where it is defined but absent in json. It could be solved manually of course, but when it comes to rich case class composition with many fields and often changes.. It might become yours nightmare ) So we decided to embed this feature directly to play-json as it could maybe help somebody. So what do you guys think about it? Is it possible?


andyglow:
I've even just found that you have corresponding issue already.. https://github.com/playframework/play-json/issues/4

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -16,9 +16,12 @@ trait PathFormat {
   def at[A](path: JsPath)(implicit f: Format[A]): OFormat[A] =
     OFormat[A](Reads.at(path)(f), Writes.at(path)(f))
 
-  def nullable[A](path: JsPath)(implicit f: Format[A]): OFormat[Option[A]] =
+   def nullable[A](path: JsPath)(implicit f: Format[A]): OFormat[Option[A]] =
     OFormat(Reads.nullable(path)(f), Writes.nullable(path)(f))
 
+  def optional[A](path: JsPath, defaultValue: => Option[A])(implicit f: Format[A]): OFormat[Option[A]] =
```
cchantep: The naming "optional" doesn't match the semantic "with-default" for me

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -52,6 +55,19 @@ trait PathReads {
     )
   }
 
+  def optional[A](path: JsPath)(implicit reads: Reads[A]) = Reads[Option[A]] { json =>
+    path.applyTillLast(json).fold(
+      jserr => jserr,
+      jsres => jsres.fold(
+        _ => JsError(path, JsonValidationError("error.path.missing")),
```
cchantep: How is this related to a support for the default values?

play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -76,6 +76,7 @@ object JsMacroImpl {
     // callNullable is the equivalent of call for options
     // e.g. `(__ \ "foo").readNullable`
     val callNullable = TermName(s"${methodName}Nullable")
```
cchantep: Better to wait the macros are refactored

gmethvin:
Could we add an option in `JsonConfiguration` to turn on/off this behavior?

Scala code isn't always meant to be translated directly to JSON validation requirements. In many cases the constructor default values are only intended for use when creating the object on the server. Also these defaults might change over time so there is some value to requiring it in validation.

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
andyglow: Well, there are two distinct cases.
We could define default value for generic field as well as define it for optional one.
Why so.. Handling reads on optional fields with default values make us think about how could we say reset optional value if there if some default.
```scala
case class Foo(bar: Option[String] = Some("bar"))
```
So. if there is no `bar` field in the payload - according to semantic we have to use default value, but is there any way to set it to None? That's a question and that's why you see this construction.

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
andyglow: Please see prev answer. Sending `null` one is still able to reset optional field with default behavour to `None`.

play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
andyglow: is there any estimation on this, guys?

andyglow:
Having this feature toggled makes sense for me. Good idea. I'll try to refactor it this way. But there is, if I got it right, still dependency on macros refactoring. Would be great to see approximate dates.

From https://docs.google.com/document/d/11sVi1-REAIDFVHvwBrfRt1uXkBzROHQYgmcZNGJtDnA/pub I see that `playframework 2.6` will be released at (ETA: March-April 2017) but what about `play-json` with it's own new release cycle?

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
cchantep: The comment was about the naming. I confirm that "optional" is unclear for me about the semantic of the function.

andyglow: I see. I thought this way. There were 2 methods - `as` and `optional`. I've just duplicated both ot them and added defaultValue param. If you look into history you will see my previous commit was done exactly using clear naming. This added a lot of boilerplate IMO. Just check

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -16,9 +16,12 @@ trait PathFormat {
   def at[A](path: JsPath)(implicit f: Format[A]): OFormat[A] =
     OFormat[A](Reads.at(path)(f), Writes.at(path)(f))
 
-  def nullable[A](path: JsPath)(implicit f: Format[A]): OFormat[Option[A]] =
+   def nullable[A](path: JsPath)(implicit f: Format[A]): OFormat[Option[A]] =
```
cchantep: Indent issue

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
cchantep: Either using previous naming, or a different one which would correspond to the semantic "nullable-with-default"

gmethvin:
I think we also should change the behavior of the `Writes` macro to always write out all the fields. For the edge case `Option[T] = Some(t)`, you need to write a `null` to represent `None` if you want to be able to read back `None` from the JSON properly. Right now nothing is written, so the default value would be used. Also if you're storing that JSON and using it with other apps, you would normally want the default values to be defined explicitly to remove any ambiguity. I suppose this could be part of the configuration.

typesafehub-validator:
Hi @andyglow,

Before we proceed with reviewing this pull request, please sign the Lightbend Contributors License Agreement:

andyglow:
Guys, thanks for your time spent on the review. I've just finished remastering where I tried to follow all the recommendation. So..
1. -with-default naming was returned
2. Writes with default was implemented (actually named -non-default)
3. For macros 'default-values' behaviour is switchable via JsonConfiguration implicit.

Please take a look

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -173,6 +190,9 @@ trait PathWrites {
   def at[A](path: JsPath)(implicit wrs: Writes[A]): OWrites[A] =
     OWrites[A] { a => JsPath.createObj(path -> wrs.writes(a)) }
 
+  def nonDefault[A](path: JsPath, defaultValue: => A)(implicit wrs: Writes[A]): OWrites[A] =
```
cchantep: What is the use case? For now I don't see a sufficiently generic case where we would like a "default" value to be written as an empty object. Moreover if `wrs` is not an `OWrites` it would lead to a possible heterogeneous JSON format (a plain value if not default, an object otherwise...). At least that must be discussed out of the scope of this PR I think.

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -186,6 +206,21 @@ trait PathWrites {
       }
     }
 
+  /**
+    * writes a optional field in given JsPath : if None, doesn't write field at all.
+    * Please note we do not write "null" but simply omit the field when None
+    * If you want to write a "null", use ConstraintWrites.optionWithNull[A]
+    */
+  def nullableNonDefault[A](path: JsPath, defaultValue: => Option[A])(implicit wrs: Writes[A]): OWrites[Option[A]] =
```
cchantep: Same as `nonDefault[A]`

play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -453,20 +461,29 @@ object JsMacroImpl {
...
         val isOption = pt.typeConstructor <:< typeOf[Option[_]].typeConstructor
-        (isOption, defaultValue, methodName) match {
-          case (true,  Some(dv), "format")          => q"OFormat($jspathTree.readOptional($impl).orElse(Reads.pure($dv)), $jspathTree.writeNullable($impl))"
-          case (true,  Some(dv), "read")            => q"$jspathTree.readOptional($impl).orElse(Reads.pure($dv))"
-          case (false, Some(dv), "format")          => q"OFormat($jspathTree.read($impl).orElse(Reads.pure($dv)), $jspathTree.write($impl))"
-          case (false, Some(dv), "read")            => q"$jspathTree.read($impl).orElse(Reads.pure($dv))"
-          case (true,  _, _)                        => q"$jspathTree.$callNullable($impl)"
-          case (false, _, _)                        => q"$jspathTree.$call($impl)"
+
+        def plain() = {
```
cchantep: `def plain = {`

play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -478,7 +495,7 @@ object JsMacroImpl {
     val applyOrMap = TermName(if (multiParam) "apply" else mapLikeMethod)
 
     val syntaxImport = if (!multiParam && !writes) q"" else q"import $syntax._"
-    @inline def buildCall = q"$canBuild.$applyOrMap(..${conditionalList(applyFunction, ApplyUnapply.unapplyFunction)})"
+    @inline def buildCall = q"""$canBuild.$applyOrMap(..${conditionalList(applyFunction, ApplyUnapply.unapplyFunction)})"""
```
cchantep: No need for the reformat

play-json/src/main/scala/play/api/libs/json/JsPath.scala
```diff
@@ -251,6 +251,11 @@ case class JsPath(path: List[PathNode] = List()) {
   /** Reads a T at JsPath */
   def read[T](implicit r: Reads[T]): Reads[T] = Reads.at[T](this)(r)
 
+  /** Reads a T at JsPath */
```
cchantep: Please indicate in the Scaladoc in which way this one differs from `read[T]`

play-json/src/main/scala/play/api/libs/json/JsPath.scala
```diff
@@ -361,6 +382,10 @@ case class JsPath(path: List[PathNode] = List()) {
 
   /** Reads/Writes a T at JsPath using provided implicit Format[T] */
   def format[T](implicit f: Format[T]): OFormat[T] = Format.at[T](this)(f)
+  /** Reads/Writes a T at JsPath using provided implicit Format[T] */
+  def formatWithDefault[T](defaultValue: => T)(implicit f: Format[T]): OFormat[T] = {
+    Format.withDefault[T](this, defaultValue)(f)
+  }
```
cchantep: Blank line after

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
andyglow: If am not mistaken this was mentioned earlier. The use cas as I could presume is optimization of traffic between `api client` and `api server`.. I agree that it's may be out of the scope.

play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
andyglow: Could you may be specify the concern? Naming? Lack of type annotation?

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
cchantep: Out of scope and to be discuss

play-json/src/main/scala/play/api/libs/json/JsMacroImpl.scala
andyglow: right. sorry

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
cchantep: Out of scope

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
andyglow: I thought about decoupling of configuration. Like 
```
case class JsonConfiguration(
  naming: JsonNaming,
  useDefaultValuesForReads: Boolean = false,
  useDefaultValuesForWrites: Boolean = false)
```
It may expand usage scenarios IMO.

cchantep:
Hi @andyglow please rebase to fix the conflict.

andyglow:
Sure. Will be back with resolved conflict and eliminated write-with-defaults behaviour in a few days. Is it ok or want it to be ready sooner?

cchantep:
As soon as you can is better I guess.

andyglow:
@cchantep, could you take a look at my changes. I've removed `-non-default` things from `write` behaviour and have rebased against recent `master` also.

play-json/src/main/scala/play/api/libs/json/JsonConfiguration.scala
```diff
@@ -3,7 +3,10 @@
  */
 package play.api.libs.json
 
-case class JsonConfiguration(naming: JsonNaming)
+case class JsonConfiguration(
+  naming: JsonNaming,
```
gmethvin: Can we define all the default values here, e.g. `naming: JsonNaming = JsonNaming.Identity`. That also means we can change the `defaultConfiguration` to just `JsonConfiguration()`.

This is nice when you want to customize the configuration from the default, since all you need to do is `JsonConfiguration(useDefaultValues = true)`

andyglow: yes. would be nice. agree

play-json/src/test/scala/play/api/libs/json/JsonExtensionSpec.scala
```diff
@@ -651,6 +654,64 @@ class JsonExtensionSpec extends Specification {
...
+      }
+
+      def macroFormat: Format[WithDefault2] = {
+        implicit val enableDefaultValues = JsonConfiguration(JsonNaming.Identity, useDefaultValues = true)
+        implicit val barFormats = Json.format[WithDefault1]
+        Json.format[WithDefault2]
+      }
+
+      def validateReads(fooReads: Reads[WithDefault2]) = {
+        fooReads.reads(Json.obj()) must beEqualTo(JsSuccess(WithDefault2()))
```
andyglow: ok, but let me ask what's wrong with `~`?

cchantep: `and` of Specs2

cchantep:
Need to be rebased

andyglow:
rebased, please check

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -52,6 +61,30 @@ trait PathReads {
     )
   }
```
cchantep: Then `nullable` there can be refactored as bellow.

```scala
def nullable[A](path: JsPath)(implicit reads: Reads[A]) =
  nullableWithDefault(path, None)
```

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -52,6 +61,30 @@ trait PathReads {
...
+   * - If one node in JsPath is not found before last node => returns JsError( "missing-path" )
+   * - If all nodes are found till last node, it runs through JsValue with last node =>
+   *   - If last node if not found => returns JsError( "missing-path" )
+   *   - If last node is found with value "null" => returns None
+   *   - If last node is found => applies implicit Reads[T]
+   */
+  def nullableWithDefault[A](path: JsPath, defaultValue: => Option[A])(implicit reads: Reads[A]) = Reads[Option[A]] { json =>
+    path.applyTillLast(json).fold(
+      jserr => jserr,
+      jsres => jsres.fold(
```
cchantep: `_.fold`

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -52,6 +61,30 @@ trait PathReads {
...
+        _ => JsSuccess(defaultValue),
+        a => a match {
```
cchantep: `_ match {`

marcospereira:
Build is broken.

@andyglow please run `sbt validateCode` to check code style. If there are any problems, the code will be reformatted and you can just commit the changes.

andyglow:
Thanks, @marcospereira. I've seen this but always forget about it. Will fix right now.

play-json/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -48,18 +48,8 @@ trait PathReads {
...
-      jsres => jsres.fold(
-        _ => JsSuccess(None),
-        a => a match {
-          case JsNull => JsSuccess(None)
-          case js => reads.reads(js).repath(path).map(Some(_))
-        }
-      )
-    )
-  }
+  def nullable[A](path: JsPath)(implicit reads: Reads[A]) =
```
cchantep: Add explicit return type

cchantep:
Hi @gmethvin (cc @schmitch ), for flag such as the `useDefaultValues` introduced there to customize the macro behaviour, I would suggest an compile-time options mechanism as [there](https://github.com/cchantep/play-json/commit/e1b1f1d6e67ca4ed967b52cce4c6aae4064a8894#diff-1c39c291029a847e81dd3b7c67cf933dR300).

That allows to manage everything at compile-time.

andyglow:
Hello, guys!
Is there anything I can help with?
@gmethvin, if I am not mistaken, your voice is blocking. So what is your opinion about the way we could toggle defaul-value behaviour? At runtime as it has already done or at compile time it preferrable? I defenitely see benefits from being able to define this behaviour at compile-time.

cchantep:
@andyglow there are remaining review comments, to be considered or discussed

andyglow:
@cchantep anything besides the way we toggle default-value behaviour?
please point me if so, can't find what I missed

cchantep:
I have indicated a commit with suggested changes in a previous comment

gmethvin:
Sorry I missed this. I've been on vacation for a few days and wasn't paying close attention.

@andyglow I think @cchantep was suggesting a refactor here: https://github.com/cchantep/play-json/commit/1e3de0ffa3744f61d88a3fbf9da738d0453feb4a

I think we should give some additional thought to the configuration strategy in general (and would be happy to have your thoughts @andyglow), but it might be best to do that as a separate pull request that changes all the configuration-related stuff. There are other options we might want to add too (e.g. property/type naming as discussed in #10). Currently `JsonConfiguration` is what we've come up with, but I can't say for sure that it's the ideal strategy for all cases. We haven't released a final version with it, so it would be fair to change it to something else.

gmethvin:
BTW I like the strategy @cchantep proposed, I just think it should be done with the naming configuration too for consistency, and any other config options we add in the future.

cchantep:
@gmethvin there are way to gather everything but I'm not sure that consistent in the end. As annotation retention, the scopes of these settings are different for me: the naming function need to be retain during runtime to be applied on runtime values, whereas the flags used by the macros are not useful at all at runtime.

andyglow:
agree with @cchantep, that configuration options which can be semantically expressed as booleans/flags could be put to `formatOpts` with all the benefits of compile-time checks and used as `formatOpts[MyCaseClass, Feature1 with Feature2]`, but if we want to provide something else there is no such a clear way for that, so all we have in the end is a configuration pased at runtime.

gmethvin:
Sure. I guess something with a custom implementation like the naming isn't that easy to represent as a type.

I'm not sure I like the name `formatOpts` though. Maybe `formatWithOpts`? And is there a way we could allow you to define a custom configuration once and use it everywhere, without explicitly passing the opts? (I guess now we're getting into the configuration discussion.)

cchantep:
I'm not fond of `formatWithOpts`.

In the draft there is a base/default type of the flags, but the type param in the macro function neither can be defaulted nor a relative implicit (no implicit other than type tags can be passed) be resolved for to manage inference.

cchantep:
@andyglow if ok for you, you could for now include the changes discussed in https://github.com/cchantep/play-json/commit/1e3de0ffa3744f61d88a3fbf9da738d0453feb4a , and rebased (not waiting the current discussion about macro flags).

andyglow:
Ok

cchantep:
Thx

andyglow:
guys, could we go with merge instead? rebase for such a big amount of changes look some overwhelming.

cchantep:
At this point this cannot be reviewed for merge without rebase. Will have a look otherwise.

cchantep:
At this point this cannot be reviewed for merge without rebase. Will have a look otherwise.

cchantep:
Ok for the cherry-pick and conflict resolution

cchantep:
cc @gmethvin Here is another draft mixing/gathering `JsonConfiguration` and compile-time flags: https://github.com/cchantep/play-json/commit/eb26ce31c1232156cd8365c416b873e8ae0c861d#diff-e45154283c526ca970477983f78729d3R43

It also [updates the naming](https://github.com/cchantep/play-json/commit/eb26ce31c1232156cd8365c416b873e8ae0c861d#diff-e45154283c526ca970477983f78729d3R302) as following when passing explicit the compile-time flags:

```scala
Json.using[Json.WithDefaultValues].reads[MyType]

// Instead of Json.readsOpts[MyType, Json.WithDefaultValues]
// in previous draft
```

gmethvin:
@cchantep I like the `Json.using` syntax. I guess that would allow you to do something like:
```scala
val configuredJson = Json.using[Json.WithDefaultValues]
configuredJson.reads[Foo]
```

Also we should think about binary compatibility, especially since Play depends on play-json. So maybe `JsonConfiguration` should be a trait instead of a case class. Or we just have an implicit `JsonNaming` dependency directly; not sure which is better design.

cchantep:
In this draft, `JsonConfiguration` is already trait, which should be ok/allow to be compatible (and more flexible).

cchantep:
@gmethvin Separate PR about the configuration/macro options: #33

cchantep:
@gmethvin Not sure about, draft about a new `JsonFacade` to gather the `Json` companion and the new `WithOptions` instances (themselves from `Json.using` or `Json.configured`), so that all the functions previously available on `Json` (not only `.reads`/`.writes`/`.format`) can be applied in the same way as bellow.

- `Json` -> `Json.{parse,stringify,...,reads,writes,format}`
- `Json.using[MacroOptions]` -> `Json.using[MacroOptions].{parse,stringify,...,reads,writes,format}`
- `Json.configured[_ <: JsonConfiguration]` -> `Json.configured.{parse,stringify,...,reads,writes,format}`

```scala
import play.api.libs.json.{ Json, JsResult }

case class Foo(bar: String = "lorem")
val input = """{"bar":"ipsum"}"""

{ // `Json` default facade
  implicit val r1: Reads[Foo] = Json.reads[Foo]
  val parsed1: JsResult[Foo] = Json.parse[Foo](input)
}

{ // Facade with explicit macro options
  val withDefaultValues = Json.using[Json.WithDefaultValues]
  implicit val r2: Reads[Foo] = withDefaultValues.reads[Foo]
  val parsed2: JsResult[Foo] = withDefaultValues.parse[Foo](input)
}

{ // Facade with macro options from the current configuration
  val configured = Json.configured
  implicit val r3: Reads[Foo] = configured.reads[Foo]
  val parsed2: JsResult[Foo] = configured.parsed[Foo](input)
}
```

> The materializers `.reads`/`.writes`/`.format` would be available on both the `Json` object and on the `WithOptions` trait, but they cannot be defined in the `JsonFacade`, as implemented by macros.

cchantep:
`JsonFacade` draft updated at https://github.com/cchantep/play-json/commit/b316c3808d15dc3ccc184e0d48400c35f49dbdff

cchantep:
Prepare the JSON facade in a separate PR #34

cchantep:
@andyglow It takes some times.

You should be able to rebase your branch/or merge the updated master, and then cherry-pick https://github.com/cchantep/play-json/commit/92c863805c13d4c9f2d8ca32d17937684d7d996f to finalize this PR.

andyglow:
ok, i'm going to prepare it today

andyglow:
Guys, i've just merged and cherry-picked, and tests went fine locally, but I haven't testeg against scala 2.10.6.. so. for now the build is broken. https://travis-ci.org/playframework/play-json/jobs/201902348
Have you already faced this?

ps: cool stuff done using typelevel programming, btw, guys!

cchantep:
@andyglow just remove `import c.universe._` as `MacroBundle` is used now

andyglow:
@cchantep done

cchantep:
Thx @andyglow

andyglow:
When should we wait for a new bright shining 2.6 version, @cchantep?

cchantep:
^_ @gmethvin

gmethvin:
M3 is out: https://oss.sonatype.org/content/repositories/releases/com/typesafe/play/play-json_2.12/2.6.0-M3/

PR #23, opened 2017-01-06
remove unused dependencies

xuwei-k:


cchantep:
@gmethvin should be ok (seems I cannot approve)

Issue #11, opened 2016-12-12
JSON readNullable erroneously fails with "error.path.empty"


gmethvin:
**(Moved from playframework/playframework#6710)**

In Play 2.5 (and current master) the JSON readers for nullable types do not work when used without a nested path.

````scala
// works
val composableRootReadsNullableScoped: Reads[Root] = (
  (__ \ "foo").read[Foo] and
  (__ \ "bar").readNullable[Bar]
)((foo, barOpt) => Root(foo, barOpt.getOrElse(Bar(-1, -1, -1))))

// does not work
val composableRootReadsNullable: Reads[Root] = (
  (__ \ "foo").read[Foo] and
   __.readNullable[Bar]
)((foo, barOpt) => Root(foo, barOpt.getOrElse(Bar(-1, -1, -1))))
````

using the following JSON inputs
````json
// for first reader
{
  "foo": { "a": 1, "b": 0 },
  "bar": {
    "barA": 0,
    "barB": 0,
    "barC": 0
  }
},

// for second reader
{
  "foo": { "a": 1, "b": 0 },
  "barA": 0,
  "barB": 0,
  "barC": 0
}
````

I expect both examples to result in the same parsed `Root` object containing the same `Foo` and `Bar`. The use-case of handling several fields with their own specialised `Reads` seems very natural to me. Often API's contain rows which are multiple merged objects like bar and foo in this example, and composing the combined `Reads` using other `Reads` is the nicest solution in my opinion.

For all types & a test case, see my full example:
https://gist.github.com/hermanbanken/98057a12f349ddbc6fab9575b8aa5f2e

````
[error] Actual:   JsError(List((,List(ValidationError(List(error.path.empty),WrappedArray())))))
[error] Expected: JsSuccess(Root(Foo(1,0),Bar(0,0,0)),)
````

The JsError originates in [JsPath.scala#L184](https://github.com/playframework/playframework/blob/2.5.x/framework/src/play-json/src/main/scala/play/api/libs/json/JsPath.scala#L184). I'm not sure, but maybe the branch checking Nil is unnecessary? Nil is a valid tail, so maybe it should be handled as a empty tail?

asheshambasta:
Just bumped into the same issue. What's the ETA on this please?

hermanbanken:
Huray 👍

PR #28, opened 2017-01-22
Fix readNullable on root path (issue #11)

pbugnion:
This PR is an attempt at addressing issue #11 .

Prior to this PR, the calling `__.readNullable[T]` would return 'error.path.missing' even when the root object was correctly formed. This commit modifies the code so that it works, at least when the object is present.

With this PR, the following works:

```scala
import play.api.libs.functional.syntax._
import play.api.libs.json._

case class User(name: String, age: Int)
case class Address(street: String, postcode: String)

implicit val readsUser: Reads[User] = (
  (__ \ "name").read[String] and (__ \ "age").read[Int]
)(User)
implicit val readsAddress: Reads[Address] = (
  (__ \ "street").read[String] and (__ \ "postcode").read[String]
)(Address)

implicit val readsUserAddress: Reads[(User, Option[Address])] = (
  __.read[User] and __.readNullable[Address]
).tupled

val sherlockHolmes = Json.obj(
  "name" -> "Sherlock Holmes",
  "age" -> 32,
  "street" -> "221B Baker Street",
  "postcode" -> "NW1 6XE"
)
sherlockHolmes.validate(readsUserAddress)
// (User("Sherlock Holmes", 32), Some(Address("221B Baker Street", "NW1 6XE")))
```

While this works, I'm not sure how to address the case where specific fields are missing. For instance, if `sherlockHolmes` just contains `User` fields:

```
val sherlockHolmes = Json.obj(
  "name" -> "Sherlock Holmes",
  "age" -> 32
)
sherlockHolmes.validate(readsUserAddress)
// JsError(List((/postcode,List(JsonValidationError(List(error.path.missing),WrappedArray()))), (/street,List(JsonValidationError(List(error.path.missing),WrappedArray())))))
```

Is this the behaviour we expect, or should this be `JsSuccess(User(...), None)`?

### DIFF
--- a/play-json/src/main/scala/play/api/libs/json/JsPath.scala
+++ b/play-json/src/main/scala/play/api/libs/json/JsPath.scala
@@ -195,7 +195,7 @@ case class JsPath(path: List[PathNode] = List()) {
 
   def applyTillLast(json: JsValue): Either[JsError, JsResult[JsValue]] = {
     def step(path: List[PathNode], json: JsValue): Either[JsError, JsResult[JsValue]] = path match {
-      case Nil => Left(JsError(Seq(this -> Seq(JsonValidationError("error.path.empty")))))
+      case Nil => Right(JsSuccess(json))
       case List(node) => node(json) match {
         case Nil => Right(JsError(Seq(this -> Seq(JsonValidationError("error.path.missing")))))
         case List(js) => Right(JsSuccess(js))
--- a/play-json/src/main/scala/play/api/libs/json/JsPath.scala
+++ b/play-json/src/main/scala/play/api/libs/json/JsPath.scala
@@ -194,6 +194,7 @@ case class JsPath(path: List[PathNode] = List()) {
   }
 
   def applyTillLast(json: JsValue): Either[JsError, JsResult[JsValue]] = {
+    @annotation.tailrec
     def step(path: List[PathNode], json: JsValue): Either[JsError, JsResult[JsValue]] = path match {
       case Nil => Right(JsSuccess(json))
       case List(node) => node(json) match {
--- a/play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
+++ b/play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
@@ -261,7 +261,7 @@ class JsonValidSpec extends Specification {
       "readNullables for missing root path fragment" >> {
         implicit val userAddressReads = (
           __.read[User] and
-            __.readNullable[Address]
+          __.readNullable[Address]
         ).tupled
 
         val missingAddressBobby = Json.obj(
@@ -271,7 +271,7 @@ class JsonValidSpec extends Specification {
 
         missingAddressBobby.validate(userAddressReads) must equalTo(
           JsError(__ \ "street", JsonValidationError("error.path.missing")) ++
-          JsError(__ \ "zip", JsonValidationError("error.path.missing"))
+            JsError(__ \ "zip", JsonValidationError("error.path.missing"))
         )
       }
 
--- a/play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
+++ b/play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
@@ -176,14 +176,14 @@ class JsonValidSpec extends Specification {
         (json.validate((__ \ "day3").read(Reads.enumNameReads(Weekdays))).asOpt must beNone)
     }
 
-    "Can reads with nullable" in {
+    "read fields with null values" in {
       val json = Json.obj("field" -> JsNull)
 
       val resultPost = json.validate((__ \ "field").read(Reads.optionWithNull[String]))
       resultPost.get must equalTo(None)
     }
 
-    "Validate options using validateOpt" in {
+    "validate options using validateOpt" in {
       val json = Json.obj("foo" -> JsNull, "bar" -> "bar")
 
       (json \ "foo").validateOpt[String] must_== JsSuccess(None)
--- a/play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
+++ b/play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
@@ -576,6 +576,18 @@ class JsonValidSpec extends Specification {
       Json.obj("type" -> "coucou", "data" -> Json.obj("title" -> "blabla")).validate(TupleReads) must beEqualTo(JsError(__ \ "data" \ "created", "error.path.missing"))
     }
 
+    "verifying reads" in {
+      case class User(id: Long, username: String)
+
+      implicit val UserReads = (
+        (__ \ 'id).read[Long] and
+        (__ \ 'username).read[String](Reads.verifying[String] { suggestedUsername => suggestedUsername.head.isLetter })
+      )(User)
+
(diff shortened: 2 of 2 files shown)


typesafehub-validator:
Hi @pbugnion,

Before we proceed with reviewing this pull request, please sign the Lightbend Contributors License Agreement:

play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
```diff
@@ -216,6 +216,49 @@ class JsonValidSpec extends Specification {
       JsString("alphabeta").validate[String] must equalTo(JsSuccess("alphabeta"))
     }
 
+    "validate reads on the root path" in {
```
cchantep: `".." >> { .. }` (No `in` in `in`)

play-json/src/main/scala/play/api/libs/json/JsPath.scala
```diff
@@ -195,7 +195,7 @@ case class JsPath(path: List[PathNode] = List()) {
 
   def applyTillLast(json: JsValue): Either[JsError, JsResult[JsValue]] = {
     def step(path: List[PathNode], json: JsValue): Either[JsError, JsResult[JsValue]] = path match {
```
cchantep: `@annotation.tailrec` could also be added

play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
pbugnion: As in, I should write:

```scala
"validate reads on the root path" in {
  "reads" >> { ... }
  "readNullables" >> { ... }
}
```

or 

```scala
"validate reads on the root path" >> {
  "reads" in { ... }
  "readNullables" in { ... }
}
```
?

play-json/src/main/scala/play/api/libs/json/JsPath.scala
pbugnion: Done

play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
cchantep: Second one

pbugnion: Done -- I've also changed some other tests in the same file that applied the same pattern.

play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
```diff
@@ -216,6 +216,49 @@ class JsonValidSpec extends Specification {
...
+        )
+      }
+
+      "readNullables" in {
+        implicit val userAddressReads: Reads[(User, Option[Address])] = (
+          __.read[User] and
+          __.readNullable[Address]
+        ).tupled
+
+        bobby.validate[(User, Option[Address])] must equalTo(
```
cchantep: Needs another test without `street` and `zip` in the JSON object, with expected `JsSuccess` with `None` address

pbugnion: I've replied as a comment below.

pbugnion:
It's not obvious (at least to me) what the correct behaviour should be when the root path is an object, but contains the wrong field. For instance, with the following reads:

```scala
case class User(name: String, age: Int)
case class Address(street: String, postcode: String)

implicit val readsUser: Reads[User] = (
  (__ \ "name").read[String] and (__ \ "age").read[Int]
)(User)
implicit val readsAddress: Reads[Address] = (
  (__ \ "street").read[String] and (__ \ "postcode").read[String]
)(Address)

implicit val readsUserAddress: Reads[(User, Option[Address])] = (
  __.read[User] and __.readNullable[Address]
).tupled
```

If the body is missing the address fields:

```scala
val sherlockHolmes = Json.obj(
  "name" -> "Sherlock Holmes",
  "age" -> 32
)
```

then: `sherlockHolmes.validate(readsUserAddress)` could be expected to return either `JsSuccess(None)` or `JsError()` corresponding to a missing address. While returning `JsSuccess(None)` seems more intuitive, it breaks some invariants. Currently (and prior to this PR), if we have:

```scala
val nestedSherlockHolmes = Json.obj("user" -> sherlockHolmes)

(nestedSherlockHolmes \ "user").validate(readsUserAddress)
// Returns JsError because the address cannot be read
```

The only time that you get `JsSuccess(None)` is if the path is missing, or the root is `JsNull`:

```scala
implicit val readsUserAddressOpts: Reads[Option[(User, Address)]] = (
  __.readNullable[User] and __.readNullable[Address]
).tupled

JsNull.validate(readsUserAddressOpts) // prior to this PR, would return JsError
// JsSuccess((None, None))

(Json.obj("user" -> JsNull) \ "user").validate(readsUserAddressOpts)
// JsSuccess((None, None))
```

cchantep:
For me, it's coherent to raise a`JsError` when reading an invalid (because empty or not containing the required field) object, wherever it is, at root or itself nested.

Your opinion @gmethvin ?

gmethvin:
There are two cases:
 1. The value to read is null or missing (e.g. the field is missing), so we return `JsSuccess(None)`
 2. The value exists, so we validate it and return the result of that validation as `JsSuccess(Some(result))` if successful, or the `JsError` if it failed.

So if we apply a `readNullable` to an invalid value (that is non-null or non-empty), it should return `JsError`.

cchantep:
So we agree the test to be added there must check the appropriate `JsError` is returned.

gmethvin:
correct

pbugnion:
Thanks a lot for the feedback. I'll add a test.

pbugnion:
I've added two unit tests, one to validate reading from a null root path and one from a badly formed root path.

play-json/src/test/scala/play/api/libs/json/JsonValidSpec.scala
```diff
@@ -216,6 +216,77 @@ class JsonValidSpec extends Specification {
...
+          "age" -> 54,
+          "street" -> "13 Main St"
+        )
+
+        missingZipBobby.validate(userAddressReads) must equalTo(
+          JsError(__ \ "zip", JsonValidationError("error.path.missing"))
+        )
+      }
+
+      "readNullables for null root path" >> {
```
cchantep: This test with `JsNull` can be kept, but I had in mind a test with a non-null/non-empty root but with no field for address (to ensure it's considered in the same way as an empty object.

pbugnion: The commit adds *two* tests. Besides the `JsNull` one, it adds:

```scala
"readNullables for badly formed root path" >> {
  implicit val userAddressReads: Reads[(User, Option[Address])] = (
    __.read[User] and
    __.readNullable[Address]
  ).tupled

  val missingZipBobby = Json.obj(
    "name" -> "bobby",
    "age" -> 54,
    "street" -> "13 Main St"
  )]
}
```

cchantep: I means a test with `Json.obj("name" -> "bobby", "age" -> 54)` ("no field for address"), which corresponds to test validating with an empty object (excluding fields for user) rather than validating an invalid object (with street but missing zip).

pbugnion: Ah that makes sense. I'll add that test.

pbugnion: I've added a unit test.

PR #32, opened 2017-01-31
Bring back 2.10 support

dwijnand:
Alternative to #26, /cc @wsargent


dwijnand:
Note to self: add 2.10 to the build

cchantep:
Guess `.travis.yml` needs also to be updated

dwijnand:
Yep, meant that too.

dwijnand:
Green.

Issue #7, opened 2016-12-12
Customization for Json macros


gmethvin:
**(Moved from playframework/playframework#4684)**

As more people start using the json macros, a lot of people seem to want to customize features slightly. We should come up with a general strategy for doing this. Ideally it should be possible to use several of these customizations at once.

Here are a few features we might want to customize:

- [X] Property naming strategy (see playframework/playframework#2333)
- [x] Using default values in `Reads` implementation (see #4)
- [ ] Serializing properties for getters (see #3306)
- [x] Renaming/ignoring specific properties (see #367)

When it comes to the behavior for things like property naming and default values, it should also be easy to use the same conventions across your application.

roti:
One need I have is a reader which handles `JsDefined(null)` and `JsUndefined` differently. For example: `
```scala
case class Xyz(foo: Int, bar: Option[String], baz: Option[Int])
implicit val xyzReader = Json.reads[Xyz]
val json = Json.parse("""{"foo": 123, "bar": null}""")

val result = json.as[Xyz] 
//expected result is Xyz(123, Some(null), None)
```
Using `Some(null)` may not be best choice to represent a json null value, but I just use it as an example here.

To my knowlegde this is not currently possible, but I may be wrong. Could I use some of the macro infrastructure from the current play release, to write such a macro myself?

Ophirr33:
+1 for allowing the usage of default values in Reads impls, specifically in the case for deserializing non-existent collections as an empty of that collection. E.g., would love to replace a lot of `Option[Seq[_]]` in our code base with just `Seq[_]`.

solarmosaic-kflorence:
I am also interested in doing with @roti mentioned above. This is especially important with HTTP `PATCH` requests, so that the user can provide intent on what they are doing ("please unset this value" vs "do not change this value"). This intent requires three states (omitted, `null` or a value) instead of lumping omitted/`null` into a single state. In other words, something like `Option[Option[T]]` (or some other model which can denote the third state) instead of `Option[T]`.

I have opened a discussion about this here: https://discuss.lightbend.com/t/distinguishing-between-omitted-and-null-values/7105

PR #33, opened 2017-02-05
Compile-time flags to customize macro usage

cchantep:
Useful  for #7 (including the pending #17)

### DIFF
--- a/play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
@@ -15,7 +15,10 @@ import java.time.{
   ZonedDateTime
 }
 import java.time.format.{ DateTimeFormatter, DateTimeParseException }
-import java.time.temporal.UnsupportedTemporalTypeException
+import java.time.temporal.{
+  Temporal => JTemporal,
+  UnsupportedTemporalTypeException
+}
 
 import com.fasterxml.jackson.databind.JsonNode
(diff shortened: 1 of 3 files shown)


schmitch:
Actually to grab the example from @gmethvin: 

```scala
val configuredJson = Json.using[Json.WithDefaultValues]
configuredJson.reads[Foo]
```
I actually think that me might look into the `using` syntax. Since it allows to create custom Json objects:

```scala
object MyJson {
private val configuredJson = Json.using[Json.WithDefaultValues]
def parse(s: String) : JsValue = configuredJson.parse(s)
}
```

which might look **way** more verbose, but we could make it simple be creating a trait which defines **all** methods and takes a `protected def configuredJson`, so that it is easy to create.

Currently we have something like that in our objects which is really error prone since especially IntelliJ not always things that the implicit is in use:

```scala
  import envisia.base.ErpConfiguration.jsonConfiguration
  import envisia.utils.date.germanDateWrites

  implicit val correctionWrites: Writes[Correction] = Json.writes[Correction]
```

Actually `ErpConfiguration.jsonConfiguration` is of type `JsonConfiguration`.
However it's really simple to remove the import on misconfigured Editors by just using "Format and Optimize Imports".

cchantep:
@schmitch As indicated in the PR #17 ( https://github.com/playframework/play-json/pull/17#issuecomment-277536100 ), what's possible (but separate from this current PR focused on the unification of compile-time flags & configuration) is to introduce a new `JsonFacade` implemented by both the current/default `Json` object and the instances from `.using` or `.configured`.

cchantep:
Thx

PR #34, opened 2017-02-07
New JsonFacade

cchantep:
See #33 and #17 : New `JsonFacade` to gather either the default `Json` object or some configured instances (see `WithOptions` trait).

### DIFF
--- a/play-json/shared/src/main/scala/JsMacroImpl.scala
+++ b/play-json/shared/src/main/scala/JsMacroImpl.scala
@@ -25,16 +25,16 @@ import scala.reflect.macros.blackbox
     }
   }
 
-  def formatImpl[A: c.WeakTypeTag]: c.Expr[OFormat[A]] =
-    macroImpl[A, OFormat, Format](
+  def formatImpl[A: c.WeakTypeTag, O <: Json.MacroOptions]: c.Expr[OFormat[A]] =
+    macroImpl[A, OFormat, Format, O](
       "format", "inmap", reads = true, writes = true
     )
 
-  def readsImpl[A: c.WeakTypeTag]: c.Expr[Reads[A]] =
-    macroImpl[A, Reads, Reads]("read", "map", reads = true, writes = false)
+  def readsImpl[A: c.WeakTypeTag, O <: Json.MacroOptions]: c.Expr[Reads[A]] =
+    macroImpl[A, Reads, Reads, O]("read", "map", reads = true, writes = false)
 
-  def writesImpl[A: c.WeakTypeTag]: c.Expr[OWrites[A]] =
-    macroImpl[A, OWrites, Writes](
+  def writesImpl[A: c.WeakTypeTag, O <: Json.MacroOptions]: c.Expr[OWrites[A]] =
+    macroImpl[A, OWrites, Writes, O](
       "write", "contramap", reads = false, writes = true
     )
 
@@ -56,7 +56,7 @@ import scala.reflect.macros.blackbox
    * @param matag The class of the reads/writes/format.
    * @param natag The class of the reads/writes/format.
    */
-  private def macroImpl[A, M[_], N[_]](methodName: String, mapLikeMethod: String, reads: Boolean, writes: Boolean)(implicit atag: c.WeakTypeTag[A], matag: c.WeakTypeTag[M[A]], natag: c.WeakTypeTag[N[A]]): c.Expr[M[A]] = {
+  private def macroImpl[A, M[_], N[_], O <: Json.MacroOptions](methodName: String, mapLikeMethod: String, reads: Boolean, writes: Boolean)(implicit atag: c.WeakTypeTag[A], matag: c.WeakTypeTag[M[A]], natag: c.WeakTypeTag[N[A]]): c.Expr[M[A]] = {
 
     // All these can be sort of thought as imports
     // that can then be used later in quasi quote interpolation
--- a/play-json/shared/src/main/scala/Json.scala
+++ b/play-json/shared/src/main/scala/Json.scala
@@ -6,39 +6,38 @@ package play.api.libs.json
 import java.io.InputStream
 
 /**
- * Helper functions to handle JsValues.
+ * @define jsonParam @param json the JsValue to convert
+ * @define returnStringRepr A String with the json representation
+ * @define readDescription Creates a `Reads[T]` by resolving, at compile-time, the case class fields or sealed family, and the required implicits.
+ * @define handlerTypeParam @tparam A the type for which the handler must be materialized
  */
-object Json {
-
+sealed trait JsonFacade {
   /**
-   * Parse a String representing a json, and return it as a JsValue.
+   * Parses a String representing a JSON input, and returns it as a [[JsValue]].
    *
-   * @param input a String to parse
-   * @return the JsValue representing the string
+   * @param input the String to parse
    */
-  def parse(input: String): JsValue = StaticBinding.parseJsValue(input)
+  def parse(input: String): JsValue
 
   /**
-   * Parse an InputStream representing a json, and return it as a JsValue.
+   * Parses a stream representing a JSON input, and returns it as a [[JsValue]].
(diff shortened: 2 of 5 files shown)


play-json/shared/src/main/scala/JsMacroImpl.scala
```diff
@@ -25,16 +25,16 @@ import scala.reflect.macros.blackbox
     }
   }
 
-  def formatImpl[A: c.WeakTypeTag]: c.Expr[OFormat[A]] =
-    macroImpl[A, OFormat, Format](
+  def formatImpl[A: c.WeakTypeTag, O <: Json.MacroOptions]: c.Expr[OFormat[A]] =
```
cchantep: Prepare the usage of the compile-time flags for the macros (see #17 ).

play-json/shared/src/main/scala/Json.scala
```diff
@@ -6,39 +6,38 @@ package play.api.libs.json
 import java.io.InputStream
 
 /**
- * Helper functions to handle JsValues.
```
cchantep: Refactor/update Scaladoc

play-json/shared/src/main/scala/Json.scala
```diff
@@ -113,23 +109,71 @@ object Json {
...
+}
+
+/**
+ * Helper functions to handle JsValues.
+ *
+ * @define macroOptions @tparam O the compile-time options
+ * @define macroWarning If any missing implicit is discovered, compiler will break with corresponding error.
+ * @define formatDescription Creates a `OFormat[T]` by resolving, at compile-time, the case class fields or sealed family, and the required implicits.
+ */
+object Json extends JsonFacade {
```
cchantep: Keep `Json`, also as the default implementation of `JsonFacade` (+ macros)

play-json/shared/src/main/scala/Json.scala
```diff
@@ -159,39 +203,92 @@ object Json {
...
+   *    (__ \ implicitly[JsonConfiguration].naming("userName")).read[String] and
+   *    (__ \ implicitly[JsonConfiguration].naming("age")).read[Int]
+   * )(User)
    * }}}
    */
-  def reads[A]: Reads[A] = macro JsMacroImpl.readsImpl[A]
+  def reads[A]: Reads[A] = macro JsMacroImpl.readsImpl[A, MacroOptions]
+
+  /** JSON facade with some macro options. */
+  final class WithOptions[Opts <: MacroOptions]() extends JsonFacade {
```
cchantep: Facade implementation `WithOptions`

play-json/shared/src/main/scala/Json.scala
```diff
@@ -159,39 +203,92 @@ object Json {
...
+     *   Json.using[Json.MacroOptions with Json.DefaultValues].reads[User]
+     */
+    def reads[A]: Reads[A] = macro JsMacroImpl.readsImpl[A, Opts]
+  }
+
+  /**
+   * Returns an inference context to call the JSON macros,
+   * using the current JSON configuration.
+   */
+  def configured[C <: JsonConfiguration.Aux[_ <: MacroOptions]](implicit config: C) = new WithOptions[config.Opts]()
```
cchantep: See https://github.com/playframework/play-json/pull/17#issuecomment-277536100

play-json/shared/src/test/scala/JsonSharedSpec.scala
```diff
@@ -26,197 +28,201 @@ class JsonSharedSpec extends WordSpec with MustMatchers {
     (__ \ 'models).format[Map[String, String]]
   )(Car, unlift(Car.unapply))
 
+  def json[T](f: JsonFacade => T) = forAll(
```
cchantep: Update tests to ensure the behaviour coherency accross the different kind of `JsonFacade`

play-json/shared/src/main/scala/Json.scala
```diff
@@ -113,23 +109,71 @@ object Json {
...
+/**
+ * Helper functions to handle JsValues.
+ *
+ * @define macroOptions @tparam O the compile-time options
+ * @define macroWarning If any missing implicit is discovered, compiler will break with corresponding error.
+ * @define formatDescription Creates a `OFormat[T]` by resolving, at compile-time, the case class fields or sealed family, and the required implicits.
+ */
+object Json extends JsonFacade {
+
+  def parse(input: String): JsValue = StaticBinding.parseJsValue(input)
```
gmethvin: Couldn't we just implement these methods in `JsonFacade` and directly call `StaticBinding` there?

play-json/shared/src/main/scala/Json.scala
```diff
@@ -6,39 +6,38 @@ package play.api.libs.json
...
 /**
- * Helper functions to handle JsValues.
+ * @define jsonParam @param json the JsValue to convert
+ * @define returnStringRepr A String with the json representation
+ * @define readDescription Creates a `Reads[T]` by resolving, at compile-time, the case class fields or sealed family, and the required implicits.
+ * @define handlerTypeParam @tparam A the type for which the handler must be materialized
  */
-object Json {
-
+sealed trait JsonFacade {
```
gmethvin: Would it be useful for this to be non-sealed? I could imagine it being useful to create a custom `JsonFacade` with special behavior (e.g. logging parsed values). Though I guess there are also risks to that.

schmitch: The question is always do the risks outweight the benefits?
I mean creating his own `JsonFacade` might be useful, so that you don't need to call `Json.using` more than once and use our own `ApplicationJson`. Tought not sure how useful that might be, it heavily depends on your second comment if we can have the `StaticBinding.methods` inside this trait.

gmethvin: I agree we should be cautious about this; just bringing up the idea. The safe and conservative thing to do is keep it sealed.

play-json/shared/src/main/scala/Json.scala
cchantep: I'd prefer not as it allows to implement that according the specificity of each `JsonFacade` (e.g. according compile-time flags possibly)

play-json/shared/src/main/scala/Json.scala
cchantep: Would also keep it sealed for now

play-json/shared/src/main/scala/Json.scala
gmethvin: Sure. I just thought it was weird to call `Json` object methods there, but looks like you fixed it. Probably makes more sense to keep `JsonFacade` methods abstract.

cchantep:
@gmethvin More tests updated, would go on with the tests in #17

cchantep:
@gmethvin Is there anything more to do?

play-json/shared/src/main/scala/JsMacroImpl.scala
jroper: One problem with this is that now there are two different ways to pass `JsonConfiguration` (using `Json.configure`, or just defining an implicit `JsonConfiguration` in the macros scope), however, only one of them allows passing macro options.  So if I do this:

```scala
implicit val myConfiguration = JsonConfiguration[WithDefaultValues](SnakeCase)
implicit val fooReads: Reads[Foo] = Json.reads[Foo]
```

I will find the surprising behavior that my reads will use snake case, however they won't use default values, because `Json.reads` is hard coded to use the default options. Is it not possible to extract the options out of the implicitly passed JsonConfiguration type member?

cchantep: Goal was to be able to define both the compile and runtime config/options, there it's prevented by not being able to pass the implicit through macros.

jroper: You can make the implicit available as a property of the same type that defines the macro (ie, make it a val on `WithOptions`), and then you can access it via `c.prefix`. See #85 :)

PR #38, opened 2017-02-15
Fix Scaladoc on JSON facades

cchantep:


play-json/shared/src/main/scala/Json.scala
```diff
@@ -211,7 +209,8 @@ object Json extends JsonFacade {
   import language.experimental.macros
 
   /**
-   * $readsDescription
```
cchantep: Scaladoc define is not working in this case of inner classes

play-json/shared/src/main/scala/Json.scala
```diff
@@ -234,7 +233,8 @@ object Json extends JsonFacade {
   def reads[A]: Reads[A] = macro JsMacroImpl.readsImpl[A, MacroOptions]
 
   /**
-   * $writesDescription
```
cchantep: Scaladoc define is not working in this case of inner classes

cchantep:
Updated

PR #40, opened 2017-02-23
Codec for Duration

cchantep:
Provides `Reads` and `Writes` for Scala durations.

*Writes[Duration]:*

| Duration                     | Json                              |
|---------------------|------------------------|
| Duration.Inf               | JsString("Inf")               |
| Duration.MinusInf     | JsString("MinusInf")     |
| Duration.Zero           | JsNumber(0)                  |
| Duration.Undefined | JsString("Undefined")    |
| _Any finite_               | JsString(finite.toString) |

*Reads[Duration]:*

| Duration                     | Json                              |
|---------------------|------------------------|
| JsString("Inf")      | Duration.Inf       |
| JsString("PlusInf")  | Duration.Inf       |
| JsString("+Inf")     | Duration.Inf       |
| JsString("MinusInf") | Duration.MinusInf |
| JsString("Undefined") | Duration.Undefined |
| JsString("1 second") | Duration("1s") |
| JsNumber(1.234D) | Duration("1234ms") |
| JsNumber(1234) | Duration("1234ms")

*Reads[FiniteDuration]:*

| Duration                     | Json                              |
|---------------------|------------------------|
| JsString("1 second") | Duration("1s") |
| JsNumber(1.234D) | Duration("1234ms") |
| JsNumber(1234) | Duration("1234ms") |

### DIFF
--- a/play-json/shared/src/main/scala/Reads.scala
+++ b/play-json/shared/src/main/scala/Reads.scala
@@ -447,7 +447,7 @@ trait DefaultReads extends LowPriorityDefaultReads {
       (n.toDouble * 1000).toLong
     } else n.toLong
 
-    FiniteDuration(millis.toLong, "ms")
+    FiniteDuration(millis, "ms")
   }
 
   /**
--- a/play-json/shared/src/main/scala/Reads.scala
+++ b/play-json/shared/src/main/scala/Reads.scala
@@ -3,6 +3,7 @@
  */
 package play.api.libs.json
 
+import scala.concurrent.duration.{ Duration, FiniteDuration }
 import scala.annotation.implicitNotFound
 
 import scala.collection._
@@ -10,8 +11,7 @@ import scala.collection.immutable.Map
 import scala.collection.mutable.Builder
 
 import scala.language.higherKinds
-
-import reflect.ClassTag
+import scala.reflect.ClassTag
 
 /**
  * Json deserializer: write an implicit to define a deserializer for any type.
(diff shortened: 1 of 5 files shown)


gmethvin:
For `FiniteDuration`s we could follow ISO8601: https://en.wikipedia.org/wiki/ISO_8601#Durations. I suspect in most cases writing to JSON your duration is finite. We could also do the same for `java.time.Period`. This way the person parsing the JSON can use the standard libraries in their language of choice. Like I said before, the goal for the default formats should be interoperability.

I think `Inf`, `MinusInf` and `Undefined` are okay but it would be nice if there was some kind of standard so you didn't have to do manual string parsing.

Thoughts?

cchantep:
This PR is following the Scala documentation about string representation of `Duration`.

I think ISO parsing better fits Java `Period`, for which such string format should already be available.

play-json/shared/src/main/scala/Reads.scala
```diff
@@ -398,8 +398,57 @@ trait DefaultReads extends LowPriorityDefaultReads {
...
    * Deserializer for Array[T] types.
    */
-  implicit def ArrayReads[T: Reads: ClassTag]: Reads[Array[T]] = new Reads[Array[T]] {
-    def reads(json: JsValue) = json.validate[List[T]].map(_.toArray)
+  implicit def ArrayReads[T: Reads: ClassTag]: Reads[Array[T]] =
+    Reads[Array[T]] { _.validate[List[T]].map(_.toArray) }
+
+  // If the number is decimal, considering it's /1000 of millis,
+  // otherwise considering it's directly in millis
+  private def numberDuration(n: BigDecimal): FiniteDuration = {
```
gmethvin: It would be nice to separate the number stuff to something like `millisDurationReads` and `millisDurationWrites`. If you want to support both in the `Reads` you could use `orElse`

play-json/shared/src/main/scala/Reads.scala
```diff
@@ -398,8 +398,57 @@ trait DefaultReads extends LowPriorityDefaultReads {
...
+    FiniteDuration(millis, "ms")
+  }
+
+  /**
+   * Deserializer for Scala FiniteDuration
+   */
+  implicit val finiteDurationWrites: Reads[FiniteDuration] =
+    Reads[FiniteDuration] {
+      case JsNumber(n) => JsSuccess(numberDuration(n))
+      case JsString("0") => JsSuccess(Duration.Zero)
```
gmethvin: This doesn't need to be a special case. Let's just write with the units "0 days".

play-json/shared/src/main/scala/Reads.scala
```diff
@@ -398,8 +398,57 @@ trait DefaultReads extends LowPriorityDefaultReads {
...
+      case JsString("Undefined") => JsError("error.invalid.finiteDuration")
+      case JsString(input) => try {
+        val d = Duration(input)
+
+        if (d.isFinite) JsSuccess(FiniteDuration(d.length, d.unit))
+        else JsError("error.invalid.finiteDuration")
+      } catch {
+        case _: Throwable => JsError("error.invalid.duration")
```
gmethvin: Don't catch all throwables here. I think `NumberFormatException` is the only thing we need to worry about.

play-json/shared/src/main/scala/Reads.scala
```diff
@@ -398,8 +398,57 @@ trait DefaultReads extends LowPriorityDefaultReads {
...
+      }
+
+      case _ => JsError("error.expected.duration")
+    }
+
+  /**
+   * Deserializer for Scala Duration
+   */
+  implicit val durationReads: Reads[Duration] = Reads[Duration] {
```
gmethvin: Can we have a `finiteDurationReads` too?

play-json/shared/src/main/scala/Writes.scala
```diff
@@ -243,6 +244,20 @@ trait DefaultWrites extends LowPriorityWrites {
   }
 
   /**
+   * Serializer for Scala Duration
+   */
+  implicit val durationWrites: Writes[Duration] = Writes[Duration] {
+    case Duration.Inf => JsString("Inf")
+    case Duration.MinusInf => JsString("MinusInf")
```
gmethvin: Maybe we should go with `-Inf`? Purely a matter of taste though.

play-json/shared/src/main/scala/Reads.scala
cchantep: 0 days is not coherent for me

play-json/shared/src/main/scala/Reads.scala
cchantep: Even if only `NumberFormatException` should be possible for now, for me that's safe to handle any exception there.

play-json/shared/src/main/scala/Writes.scala
cchantep: `Inf` and `MinusInf` are the case name in `Duration`, so for me that's less suprising

play-json/shared/src/main/scala/Reads.scala
cchantep: Already provided

play-json/shared/src/main/scala/Reads.scala
cchantep: Updated with `finiteDurationNumberReads` and `finiteDurationNumberWrites`

play-json/shared/src/main/scala/Reads.scala
cchantep: Updated as it's supported by the `JsNumber` support

cchantep:
Updated

play-json/shared/src/main/scala/Reads.scala
```diff
@@ -373,34 +379,89 @@ trait DefaultReads extends LowPriorityDefaultReads {
...
-  }
+  implicit def ArrayReads[T: Reads: ClassTag]: Reads[Array[T]] =
+    Reads[Array[T]] { _.validate[List[T]].map(_.toArray) }
+
+  private def finiteDurationNumberReads(esuffix: String) =
+    Reads[FiniteDuration] {
+      case JsNumber(n) => JsSuccess {
+        val millis = if (!n.ulp.isWhole) {
+          (n.toDouble * 1000).toLong
+        } else n.toLong
```
cchantep: It could rather be rejected as an error

gmethvin: This could also fail if the number doesn't fit in a long.

gmethvin: I'm also ok with doing some sort of rounding/truncation, but it should be consistent for all numbers.

cchantep: Updated

cchantep:
Updated

PR #41, opened 2017-02-23
Utility to convert JsResult to Try

cchantep:
```scala
def foo[T](res: JsResult[T]): Try[T] = JsResult.toTry(res)
def bar[T](res: JsResult[T]): Future[T] = Future.fromTry(JsResult.toTry(res))
```

### DIFF
--- a/play-json/shared/src/main/scala/JsResult.scala
+++ b/play-json/shared/src/main/scala/JsResult.scala
@@ -201,7 +201,7 @@ object JsResult {
   import play.api.libs.functional._
 
   case class Exception(cause: JsError)
-    extends IllegalArgumentException(Json stringify JsError.toJson(cause))
+    extends java.lang.Exception(Json stringify JsError.toJson(cause))
     with scala.util.control.NoStackTrace
 
   /**
--- a/play-json/shared/src/main/scala/JsResult.scala
+++ b/play-json/shared/src/main/scala/JsResult.scala
@@ -197,7 +197,32 @@ sealed trait JsResult[+A] { self =>
 }
 
 object JsResult {
+  import scala.util.{ Failure, Try, Success }
   import play.api.libs.functional._
+
+  case class Exception(cause: JsError)
+    extends IllegalArgumentException(Json stringify JsError.toJson(cause))
+    with scala.util.control.NoStackTrace
+
+  /**
+   * Returns a JSON validation as a [[scala.util.Try]].
+   *
+   * @tparam T the type for the parsing
+   * @param result the JSON validation result
+   * @param err the function to be applied if the results is an error
+   *
+   * {{{
+   * import scala.concurrent.Future
+   * import play.api.libs.json.JsResult
+   *
+   * def toFuture[T](res: JsResult[T]): Future[T] =
+   *   Future.fromTry(JsResult.toTry(res))
+   * }}}
+   */
+  def toTry[T](result: JsResult[T], err: JsError => Throwable = Exception(_)): Try[T] = result match {
+    case e @ JsError(_) => Failure(err(e))
+    case s @ JsSuccess(v, _) => Success(v)
+  }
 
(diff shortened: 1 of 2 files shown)


play-json/shared/src/main/scala/JsResult.scala
```diff
@@ -197,8 +197,33 @@ sealed trait JsResult[+A] { self =>
 }
 
 object JsResult {
+  import scala.util.{ Failure, Try, Success }
   import play.api.libs.functional._
 
+  case class Exception(cause: JsError)
+    extends IllegalArgumentException(Json stringify JsError.toJson(cause))
```
gmethvin: Why `IllegalArgumentException`? `RuntimeException` would probably be fine.

play-json/shared/src/main/scala/JsResult.scala
```diff
@@ -197,8 +197,33 @@ sealed trait JsResult[+A] { self =>
...
+    with scala.util.control.NoStackTrace
+
+  /**
+   * Returns a JSON validation as a [[scala.util.Try]].
+   *
+   * @tparam T the type for the parsing
+   * @param result the JSON validation result
+   * @param err the function to be applied if the results is an error
```
gmethvin: s/results/result/

play-json/shared/src/main/scala/JsResult.scala
```diff
@@ -197,8 +197,33 @@ sealed trait JsResult[+A] { self =>
 }
 
 object JsResult {
+  import scala.util.{ Failure, Try, Success }
   import play.api.libs.functional._
 
+  case class Exception(cause: JsError)
```
gmethvin: Maybe this should be in `JsError`? i.e. `JsError.Exception` is an exception produced from a `JsError`.

play-json/shared/src/main/scala/JsResult.scala
cchantep: Illegal argument makes more sense, since raised when input data cannot be validated.

play-json/shared/src/main/scala/JsResult.scala
cchantep: Updated

play-json/shared/src/main/scala/JsResult.scala
cchantep: Updated

play-json/shared/src/main/scala/JsResult.scala
gmethvin: I'm just not clear on which method is being passed an invalid argument. See https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html:
> Thrown to indicate that a method has been passed an illegal or inappropriate argument.

cchantep: `.validate` is passed an invalid JSON argument in this case

gmethvin: Normally you'd use `IllegalArgumentException` to indicate that input arguments are malformed in some way. It generally indicates a programmer error. I've never seen it used when validating user input and I don't think a typical Scala user would expect that either.

gmethvin: Also, it might be nice to add a `toJson` method on the exception to make it easy to convert to a result:
```scala
lazy val toJson: JsValue = JsError.toJson(cause)
```
`exception.toJson` is a lot clearer than `JsError.toJson(exception.cause)`.

cchantep: "typical Scala user would expect that either." That's not the point, as the Scala dev will sees it's a `JsError.Exception`, which is quite explicit for me.

The `.toJson` is not useful, as the `Exception` will be in case of convertion from JSON validation to `Try`. The `.toJson` is useful when considering the error as validation error.

cchantep:
thx

PR #42, opened 2017-02-23
Support Java LocalTime

cchantep:


play-json/jvm/src/main/scala/play/api/libs/json/EnvWrites.scala
```diff
@@ -134,14 +141,35 @@ trait EnvWrites {
...
+   * Serializer for `java.time.LocalTime` as JSON number.
+   *
+   * {{{
+   * import java.time.LocalTime
+   * import play.api.libs.json.Writes
+   *
+   * implicit val ltnWrites = Writes.LocalTimeNumberWrites
+   * }}}
+   */
+  val LocalTimeNumberWrites: Writes[LocalTime] = Writes[LocalTime] { t =>
```
gmethvin: Might be useful to be able to pass units here, e.g. to write as hours, minutes, milliseconds, etc.

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
```diff
@@ -402,6 +415,49 @@ trait EnvReads {
...
+  /**
+   * Reads for the `java.time.LocalTime` type.
+   *
+   * @tparam T Type of argument to instantiate time parser
+   * @param parsing Argument to instantiate time parser. Actually either a pattern (string) or a formatter (`java.time.format.DateTimeFormatter`)
+   * @param corrector a simple string transformation function that can be used to transform input String before parsing. Useful when standards are not exactly respected and require a few tweaks. Function `identity` can be passed if no correction is needed.
+   * @param p Typeclass instance based on `parsing`
+   * @see [[DefaultWrites.TemporalFormatter]]
+   * {{{
+   * import play.api.libs.json.Java8Reads.localTimeReads
```
gmethvin: Where is `Java8Reads` defined?

play-json/jvm/src/main/scala/play/api/libs/json/EnvWrites.scala
cchantep: The `Writes` with format is there for such usage

cchantep:
Updated

play-json/jvm/src/main/scala/play/api/libs/json/EnvWrites.scala
gmethvin: How would you do that? What I mean is you could use `localTimeNumberWrites(TimeUnit.SECONDS)` to write the number of seconds in the day as a number.

Also, this uses nanoseconds by default and the `LocalDateTimeNumberWrites` uses milliseconds since the epoch. If I were to use both together I would expect both to default to the same time unit.

cchantep: `LocalTime` doesn't use milliseconds epoch but nanos of day (not similar kind). I wouldn't find appropriate to convert this case.

The second case could be separate.

gmethvin: > LocalTime doesn't use milliseconds epoch but nanos of day

It just seemed odd to me that writing a `LocalDate` as a number would use different units than `LocalTime`.

Can we just rename the reads/writes to specify the units they are using. e.g. `LocalDateNanosOfDayWrites`, `LocalTimeMillisWrites`. I'm ok with either that or taking a `TimeUnit` as a parameter.

cchantep: Naming updated

gmethvin:
There are some scalariform issues so fix those before you merge.

gmethvin:
Also, I think `Java8Reads` refers to a class that's not here. Other than that, looks good 👍 .

cchantep:
Updated

PR #43, opened 2017-02-25
Codecs for Java TemporalAmount

cchantep:
Support Java Time `Duration` (see #40 ) and `Period`, either as number (by default of millis of `Duration` or of `days` for `Period`), or as ISO string representation.

### DIFF
--- a/play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
@@ -377,7 +377,7 @@ trait EnvReads {
         case JsNumber(d) => JsSuccess(epoch(d.toLong))
         case JsString(s) => p(parsing).parse(corrector(s)) match {
           case Some(d) => JsSuccess(d)
-          case None => JsError(Seq(JsPath ->
+          case _ => JsError(Seq(JsPath ->
             Seq(JsonValidationError("error.expected.date.isoformat", parsing))))
         }
         case _ => JsError(Seq(JsPath ->
@@ -445,7 +445,7 @@ trait EnvReads {
         case JsNumber(d) => JsSuccess(epoch(d.toLong))
         case JsString(s) => p(parsing).parse(corrector(s)) match {
           case Some(d) => JsSuccess(d)
-          case None => JsError(Seq(JsPath ->
+          case _ => JsError(Seq(JsPath ->
             Seq(JsonValidationError("error.expected.date.isoformat", parsing))))
         }
         case _ => JsError(Seq(JsPath ->
@@ -538,7 +538,7 @@ trait EnvReads {
         JsError("error.invalid.longDuration")
 
       case JsNumber(n) => JsSuccess(JDuration.of(n.toLong, unit))
-      case _ => JsError("error.expected.duration")
+      case _ => JsError("error.expected.lonDuration")
     }
 
   /**
@@ -561,7 +561,7 @@ trait EnvReads {
     case JsString(repr) => try {
       JsSuccess(JDuration.parse(repr))
     } catch {
-      case _: Exception => JsError("error.invalid.duration")
+      case _: DateTimeParseException => JsError("error.invalid.duration")
     }
 
     case js => javaDurationMillisReads.reads(js)
@@ -646,7 +646,7 @@ trait EnvReads {
     def reads(json: JsValue): JsResult[LocalDate] = json match {
       case JsString(s) => parseDate(corrector(s)) match {
         case Some(d) => JsSuccess(d)
-        case None => JsError(Seq(JsPath() -> Seq(JsonValidationError("error.expected.jodadate.format", pattern))))
+        case _ => JsError(Seq(JsPath() -> Seq(JsonValidationError("error.expected.jodadate.format", pattern))))
       }
       case _ => JsError(Seq(JsPath() -> Seq(JsonValidationError("error.expected.date"))))
     }
--- a/play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
@@ -8,10 +8,12 @@ import java.util.Locale
 import java.time.{
   Clock,
   DateTimeException,
+  Duration => JDuration,
   Instant,
   LocalDate,
   LocalTime,
   LocalDateTime,
+  Period,
   OffsetDateTime,
   ZoneId,
   ZoneOffset,
@@ -19,6 +21,8 @@ import java.time.{
 }
 import java.time.format.{ DateTimeFormatter, DateTimeParseException }
 import java.time.temporal.{
+  ChronoUnit,
+  TemporalUnit,
   Temporal => JTemporal,
   UnsupportedTemporalTypeException
 }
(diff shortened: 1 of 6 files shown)


play-json/jvm/src/main/scala/play/api/libs/json/EnvWrites.scala
```diff
@@ -207,6 +209,28 @@ trait EnvWrites {
...
+  /**
+   * Serializer of Java Duration using ISO representation
+   * (e.g. PT1S for 1 second).
+   */
+  implicit val javaDurationWrites: Writes[JDuration] =
+    Writes[JDuration] { d => JsString(d.toString) }
+
+  /** Serializer of Java Period as a number of days. */
+  val javaPeriodDaysWrites: Writes[Period] =
+    Writes[Period] { d => JsNumber(d.getDays) }
```
gmethvin: `getDays` returns only the `days` field in the `Period`. I'd think you'd want to convert the months and years to days and include those at well. You could use `Duration.from(period).toDays`.

It would also be useful to have constants for years and months like I suggested for the `Reads`

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
```diff
@@ -415,6 +419,78 @@ trait EnvReads {
     case _ => JsError(JsonValidationError("error.expected.jsstring"))
   }
 
+  private def javaDurationNumberReads(unit: TemporalUnit, esuffix: String) =
+    Reads[JDuration] {
+      case JsNumber(n) if !n.ulp.isWhole =>
+        JsError(s"error.invalid.$esuffix")
```
cchantep: I prefer to have an error key related to the duration case.

I don't think rounding should happen there. The temporal unit is specified and a corresponding integer should be found, or the validation indicates there is a misusage.

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
```diff
@@ -415,6 +419,78 @@ trait EnvReads {
...
+  /**
+   * Deserializer of Java Duration, from either a time-based amount of time
+   * (string representation such as '34.5 seconds'),
+   * or from a number of milliseconds (see [[javaDurationMillisReads]]).
+   */
+  implicit val DefaultJavaDurationReads: Reads[JDuration] = Reads[JDuration] {
+    case JsString(repr) => try {
+      JsSuccess(JDuration.parse(repr))
+    } catch {
+      case _: Throwable => JsError("error.invalid.duration")
```
cchantep: I don't see reason not to catch any exception there, even if for now only this one is supposed to happen.

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
```diff
@@ -415,6 +419,78 @@ trait EnvReads {
...
+   */
+  def javaPeriodNumberReads(unit: TemporalUnit): Reads[Period] = {
+    val underlying = javaDurationNumberReads(unit, "period")
+
+    Reads[Period] { js =>
+      underlying.reads(js).flatMap { duration =>
+        try {
+          JsSuccess(Period.ofDays(duration.toDays.toInt))
+        } catch {
+          case e: Throwable => JsError("error.invalid.period")
```
cchantep: As for `{ Future, Try }.{ map, flatMap }`, `JsResult` should be safe. An exception while parsing period should be represented with the `JsResult` not letting the exception bubbling up.

The point is there is no general mechanism about that in the `JsResult` combinators.

play-json/jvm/src/main/scala/play/api/libs/json/EnvWrites.scala
cchantep: Indeed there is no stable way to write a `Period` as number (`Duration.from(aPeriod)` is not supported as `Period` is an estimated duration).

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
```diff
@@ -415,6 +419,82 @@ trait EnvReads {
     case _ => JsError(JsonValidationError("error.expected.jsstring"))
   }
 
+  private def javaDurationNumberReads(unit: TemporalUnit, esuffix: String) =
+    Reads[JDuration] {
+      case JsNumber(n) if !n.ulp.isWhole =>
+        JsError(s"error.invalid.$esuffix")
+
+      case JsNumber(n) => JsSuccess(JDuration.of(n.toLong, unit))
```
gmethvin: You should probably return an error if the number isn't representable as long. `BigDecimal#toLong` will start providing nonsense values if the number is outside the range of a long, for example
```
scala> val longValue = BigDecimal("1E20").toLong
longValue: Long = 7766279631452241920
```

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
gmethvin: Also, you probably don't need the `esuffix` anymore since this is only used for `Duration`.

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
gmethvin: `Duration.parse` is part of the Java standard library, so it's probably safe to assume that if it only says `@throws DateTimeParseException`, that's the only thing you're expected to handle. Any other kind of exception likely has nothing to do with whether the duration is valid. I would prefer those exceptions be passed up to the caller so they can inspect the exception and figure out what to do.

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
gmethvin: > An exception while parsing period should be represented with the JsResult not letting the exception bubbling up.

I disagree. `JsResult` is not a `Try`. `JsError` really only has the capacity to represent JSON parsing and validation errors, and I'm pretty sure that's what users expect of it. By presenting any exception as if it were invalid input, you're potentially masking other more dangerous errors.

play-json/jvm/src/main/scala/play/api/libs/json/EnvWrites.scala
gmethvin: Yeah, I think we can just avoid writing periods as numbers.

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
cchantep:
```
scala> BigDecimal(10.0D).isWhole
res0: Boolean = true

scala> BigDecimal("10.0").isWhole
res1: Boolean = true
```

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
cchantep:
```
scala> BigDecimal("1E20").isValidLong
res3: Boolean = false
```

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
```diff
@@ -415,6 +419,82 @@ trait EnvReads {
...
+    Reads[Period] {
+      case JsNumber(n) if !n.isValidInt =>
+        JsError(s"error.invalid.period")
+
+      case JsNumber(n) => JsSuccess(f(n.toInt))
+      case _ => JsError(s"error.expected.period")
+    }
+
+  /** Deserializer of Java Period from a number (integer) of days. */
+  val javaPeriodDaysReads: Reads[Period] =
```
cchantep: Right

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
gmethvin: So it seems like we should be using .isValidLong instead of n.ulp.isWhole.

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
```diff
@@ -415,6 +419,82 @@ trait EnvReads {
...
+    }
+
+    case js => javaDurationMillisReads.reads(js)
+  }
+
+  // Should only be called with a unit supported by Period
+  private def javaPeriodNumberReads(f: Int => Period): Reads[Period] =
+    Reads[Period] {
+      case JsNumber(n) if !n.isValidInt =>
+        JsError(s"error.invalid.period")
```
gmethvin: Another note about message keys: I would expect the error to be different if I'm reading a number versus if I'm reading as a period string. I'd assume `error.invalid.period` would correspond to a message like "Invalid period", which is not enough information for me to understand why this is failing.

cchantep:
Updated

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
```diff
@@ -528,6 +532,72 @@ trait EnvReads {
     }
   }
 
+  private def jdurationNumberReads(unit: TemporalUnit) =
+    Reads[JDuration] {
+      case JsNumber(n) if !n.ulp.isValidLong =>
+        JsError("error.invalid.longDuration")
+
+      case JsNumber(n) => JsSuccess(JDuration.of(n.toLong, unit))
+      case _ => JsError("error.expected.duration")
```
gmethvin: `error.expected.longDuration`?

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
```diff
@@ -528,6 +532,72 @@ trait EnvReads {
...
+  /**
+   * Deserializer of Java Duration, from either a time-based amount of time
+   * (string representation such as '34.5 seconds'),
+   * or from a number of milliseconds (see [[javaDurationMillisReads]]).
+   */
+  implicit val DefaultJavaDurationReads: Reads[JDuration] = Reads[JDuration] {
+    case JsString(repr) => try {
+      JsSuccess(JDuration.parse(repr))
+    } catch {
+      case _: Exception => JsError("error.invalid.duration")
```
gmethvin: Should be `case _: DateTimeParseException` since that's the only exception that indicates an invalid duration.

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
```diff
@@ -528,6 +532,72 @@ trait EnvReads {
...
+  /**
+   * Deserializer of Java Period, from either a time-based amount of time
+   * (string representation such as '34.5 seconds'),
+   * or from a number of milliseconds (see [[javaPeriodMillisReads]]).
+   */
+  implicit val DefaultJavaPeriodReads: Reads[Period] = Reads[Period] {
+    case JsString(repr) => try {
+      JsSuccess(Period.parse(repr))
+    } catch {
+      case _: Exception => JsError("error.invalid.stringPeriod")
```
gmethvin: `DateTimeParseException`

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
cchantep: Either we catch nothing there, having to ensure there is a generic mechanism making `JsResult` safe (for me no exception must be raised from), or for now `Reads` such as this one should catch everything.

gmethvin: I would consider it unsafe to catch an unexpected exception and map it to what looks like a validation error. This hides the true cause of the error from the caller.

If code like this throws an exception we don't already know about, it's likely because of either a bug or another serious issue on the server. I don't believe these things are our job to handle. `JsResult` is not a generic error handling mechanism; it's specifically for JSON validation. If you want to handle other exceptions, you should use a `Try`. With `JsResult.toTry`, you can do:

```scala
val result: Try[Foo] = Try(Json.fromJson[Foo](json)).flatMap(JsResult.toTry)
```

If we are trying to make the API safer, I would start with methods like `Json.parse`, and make sure we at least document the exceptions thrown if the JSON is invalid. Then we can start thinking about how to provide better ways for handling exceptions in general. But I think that is best done at a high level rather than being the responsibility of each individual `Reads`.

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
cchantep: Updated

play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
cchantep: Ok let this aside from this PR.

Will try to draft something about separately.

PR #44, opened 2017-02-25
Codec for Locale

cchantep:
See playframework/playframework#7011


gmethvin:
@cchantep Maybe we should abandon playframework/playframework#7011 in favor of doing it for Locale here, and write the `Lang` formats in Play in terms of `Locale`.

cchantep:
That would require to wait a release. The other one could be refactored after json release.

gmethvin:
I'm ok with doing another milestone after this is merged, and updating the other PR.

cchantep:
thx

PR #46, opened 2017-02-26
Refactor the JsResult combinator according the JsSuccess/JsError cases

cchantep:
- Optimize the combinators according the cases
- Add Scaladoc


cchantep:
thx

PR #49, opened 2017-03-04
Start documentating the trait support

cchantep:


docs/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
```diff
@@ -34,13 +34,15 @@ The above example can be made even more concise by using body parsers with a typ
 
 ### Requirements
 
-These macros rely on a few assumptions about the type they're working with :
-- It must have a companion object having `apply` and `unapply` methods
+These macros rely on a few assumptions about when working with a class:
```
gmethvin: Wording could be more clear. I would say something like this:

```
These macros currently support `class`es and `sealed trait` hierarchies. When working with a class:
```

docs/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
```diff
@@ -34,13 +34,15 @@ The above example can be made even more concise by using body parsers with a typ
...
+
+- It must have a companion object having `apply` and `unapply` methods.
 - The return types of the `unapply` must match the argument types of the `apply` method.
 - The parameter names of the `apply` method must be the same as the property names desired in the JSON.
 
-Case classes natively meet these requirements. For more custom classes or traits, you might
-have to implement them.
+Case classes natively meet these requirements. For more custom classes or traits, you might have to implement them.
+
+A trait can also supported, if and only if it's a sealed one and if the sub-types comply with the previous requirements.
```
gmethvin: We should at least document the basic format here for sealed traits. Perhaps we should for classes as well, though it's much easier to guess what it will look like there.

docs/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
```diff
@@ -34,13 +34,15 @@ The above example can be made even more concise by using body parsers with a typ
 
 ### Requirements
 
-These macros rely on a few assumptions about the type they're working with :
-- It must have a companion object having `apply` and `unapply` methods
+These macros support classes and sealed traits. It relies on the following assumptions.
```
gmethvin: This section is a bit confusing because it does not apply to sealed traits. Might be better to say "The macros automatically work for classes meeting following requirements:"

docs/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
```diff
@@ -34,13 +34,15 @@ The above example can be made even more concise by using body parsers with a typ
...
+
+- It must have a companion object having `apply` and `unapply` methods.
 - The return types of the `unapply` must match the argument types of the `apply` method.
 - The parameter names of the `apply` method must be the same as the property names desired in the JSON.
 
-Case classes natively meet these requirements. For more custom classes or traits, you might
-have to implement them.
+Case classes automatically meet these requirements. For custom classes or traits, you might have to implement them.
+
+A trait can also supported, if and only if it's a sealed one and if the sub-types comply with the previous requirements.
```
gmethvin: The wording is a bit strange. How about "A sealed trait can be supported if all the subtypes comply with the above requirements". Then it would be nice to explain how sealed trait instances are represented in JSON.

cchantep: It would no longer emphasise that only sealed trait is supported, precision that seems important to me

docs/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
cchantep: "The macros ~~automatically~~ for classes and traits meeting the following requirements" ?

gmethvin: "The macros work for classes and traits meeting the following requirements"

docs/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
gmethvin: Aren't unsealed traits supported if they have matching apply/unapply in their companion?

cchantep: That's the previous case for me ("For custom classes or traits, you might have to implement them").

There is more about sealed family.

gmethvin: That's why I thought saying "A sealed trait is supported if ..." is more clear than "A trait can be supported if and only if ...".

cchantep: I find clearer to indicate that `trait` if complying with the 2 conditions (being sealed, and compliant sub-types), rather than "jumping" directly to `sealed trait`.

I mean, being `sealed` and having compliant sub-types are conditions of the same importance, to be considered for the `trait` concept.

cchantep:
up

cchantep:
up

PR #54, opened 2017-03-17
Add Scala 2.10 cross build

marcospereira:
## Purpose

We were not publishing/build for Scala 2.10. This adds Scala 2.10 to cross Scala versions.

## References

See #32 where Scala 2.10 support was introduced.


marcospereira:
This has a green build now.

PR #57, opened 2017-03-24
extends Function1 for convenience and 2.5.x compatibility

xuwei-k:
2.5.x

```scala
val foo: Option[Seq[JsValue]] = None
foo.map(JsArray) // success
```

2.6.0-M6

```scala
val foo: Option[Seq[JsValue]] = None
foo.map(JsArray) // compile error
foo.map(JsArray(_)) // need (_)
```

http://stackoverflow.com/a/3054165


gmethvin:
@xuwei-k I'm ok with adding the functionality, but I'd prefer if we extend `Function1` instead of `AbstractFunction1` since `scala.runtime` is considered internal.

PR #65, opened 2017-05-12
Standardize/Fix behavior of JsLookup#apply and JsLookup#\

lihaoyi:
# Pull Request Checklist

## Fixes

I don't see an open issue for this

## Purpose

- Fix behavior of `JsLookup#apply(index: Int)` to properly throw exceptions, as is already described in the readme.

- Add a `JsLookup#apply(fieldName: String)` which similarly throws exceptions if the key is not found, for consistency with `JsLookup#apply(index: Int)` and `JsLookup#\(index: Int)` and `JsLookup#\(fieldName: String)`

- Convert an internal callsite from `JsLookup#apply(index: Int)` to `JsLookup#\(index: Int)`, since that callsite seems to want the don't-throw-exception behavior

- Document behavior of existing `JsLookup#\(index: Int)` and new `JsLookup#apply(fieldName: String)` in the readme

- Add tests for:

    - The existing `JsLookup#apply(index: Int)` to ensure it throws exceptions as is described in the readme,
    - The new `JsLookup#apply(fieldName: String)` operation

## Background Context

- `JsLookup#apply(index: Int)` was a simple alias for `JsLookup#\(index: Int)`, which meant it wasn't throwing which is what the existing readme said it should be doing

- If we have both exception-throwing and non-exception-throwing ways of looking up `JsArrays` by `index: Int`, it makes sense to have the same two ways of looking up `JsObjects` by `fieldName: String`

I haven't discussed this with anyone, but given the existing code and the existing behavior the readme describes, this seems like the right thing to do. Let me know what you think...

## References

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: scala
 scala:
-- 2.12.1
-- 2.11.8
+- 2.12.2
+- 2.11.11
 - 2.10.6
 jdk:
 - oraclejdk8
--- a/build.sbt
+++ b/build.sbt
@@ -13,12 +13,12 @@ import com.typesafe.tools.mima.plugin.MimaKeys.{
 
 resolvers ++= DefaultOptions.resolvers(snapshot = true)
 
-val specsVersion = "3.8.6"
+val specsVersion = "3.8.9"
 val specsBuild = Seq(
   "specs2-core"
 ).map("org.specs2" %% _ % specsVersion)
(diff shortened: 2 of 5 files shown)


lihaoyi:
Huh, not sure why it closed itself, let me re-open it

lihaoyi:
https://github.com/playframework/play-json/pull/66

PR #66, opened 2017-05-12
Standardize/Fix behavior of JsLookup#apply and JsLookup#\

lihaoyi:
# Pull Request Checklist

## Fixes

I don't see an open issue for this

## Purpose

- Fix behavior of `JsLookup#apply(index: Int)` to properly throw exceptions, as is already described in the readme.

- Add a `JsLookup#apply(fieldName: String)` which similarly throws exceptions if the key is not found, for consistency with `JsLookup#apply(index: Int)` and `JsLookup#\(index: Int)` and `JsLookup#\(fieldName: String)`

- Convert an internal callsite from `JsLookup#apply(index: Int)` to `JsLookup#\(index: Int)`, since that callsite seems to want the don't-throw-exception behavior

- Document behavior of existing `JsLookup#\(index: Int)` and new `JsLookup#apply(fieldName: String)` in the readme

- Add tests for:

    - The existing `JsLookup#apply(index: Int)` to ensure it throws exceptions as is described in the readme,
    - The new `JsLookup#apply(fieldName: String)` operation

## Background Context

- `JsLookup#apply(index: Int)` was a simple alias for `JsLookup#\(index: Int)`, which meant it wasn't throwing which is what the existing readme said it should be doing

- If we have both exception-throwing and non-exception-throwing ways of looking up `JsArrays` by `index: Int`, it makes sense to have the same two ways of looking up `JsObjects` by `fieldName: String`

I haven't discussed this with anyone, but given the existing code and the existing behavior the readme describes, this seems like the right thing to do. Tests seem to pass, though the old incorrect-according-to-readme behavior of `JsLookup#apply(index: Int)` didn't appear to be tested anywhere so the change in semantics didn't break anything until I added my own tests.

Let me know what you think...

## References

### DIFF
--- a/docs/manual/working/scalaGuide/main/json/ScalaJson.md
+++ b/docs/manual/working/scalaGuide/main/json/ScalaJson.md
@@ -101,9 +101,14 @@ You can traverse a `JsValue` structure and extract specific values. The syntax a
 
 ### Simple path `\`
 
-Applying the `\` operator to a `JsValue` will return the property corresponding to the field argument, supposing this is a `JsObject`.
+Applying the `\` operator to a `JsValue` will return the property corresponding to the field argument in a `JsObject`, or the item at that index in a `JsArray`
 
 @[traverse-simple-path](code/ScalaJsonSpec.scala)
+
+
+The `\` operator returns a `JsLookupResult`, which is either `JsDefined` or `JsUndefined`. You can chain multiple `\` operators, and the result will be `JsUndefined` if any intermediate value cannot be found. Calling `get` on a `JsLookupResult` attempts to get the value if it is defined and throws an exception if it is not.
+
+You can also use the Direct lookup `apply` method (below) to get a field in an object or index in an array. Like `get`, this method will throw an exception if the value does not exist.
 
 ### Recursive path `\\`
 
@@ -111,11 +116,13 @@ Applying the `\\` operator will do a lookup for the field in the current object 
 
 @[traverse-recursive-path](code/ScalaJsonSpec.scala)
 
-### Index lookup (for JsArrays)
+### Direct lookup
 
-You can retrieve a value in a `JsArray` using an apply operator with the index number.
+You can retrieve a value in a `JsArray` or `JsObject` using an `.apply` operator, which is identical to the Simple path `\` operator except it returns the value directly (rather than wrapping it in a `JsLookupResult`) and throws an exception if the index or key is not found:
 
 @[traverse-array-index](code/ScalaJsonSpec.scala)
+
+This is useful if you are writing quick-and-dirty code and are accessing some JSON values you *know* to exist, for example in one-off scripts or in the REPL.
 
 ## Converting from a JsValue
 
--- a/README.md
+++ b/README.md
@@ -73,6 +73,13 @@ val lat = (json \ "location" \ "lat").get
 
 The `(json \ "location" \ "lat")` returns a `JsLookupResult` which may or may not contain a value. Note that the `get` operation is not always safe; it throws an exception if the path doesn't exist.
 
+You can also use `\` to look up indices within a `JsArray`:
+
+```scala
+val bigwig = (json \ "residents" \ 1).get
+// returns {"name":"Bigwig","age":6,"role":"Owsla"}
+```
+
 ### Recursive path \\
 Applying the `\\` operator will do a lookup for the field in the current object and all descendants.
 
@@ -81,15 +88,18 @@ val names = json \\ "name"
 // returns Seq(JsString("Watership Down"), JsString("Fiver"), JsString("Bigwig"))
 ```
 
-### Index lookup (for JsArrays)
-You can retrieve a value in a JsArray using an apply operator with the index number.
+### Index lookup
+You can retrieve a value in a JsObject or JsArray using an apply operator with the index number or key.
 
 ```scala
-val bigwig = (json \ "residents")(1)
+val name = json("name")
+// returns JsString("Watership Down")
+
+val bigwig = json("residents")(1)
 // returns {"name":"Bigwig","age":6,"role":"Owsla"}
 ```
 
-Like `get`, this operation will throw an exception if the index doesn't exist. Use `validate` or one of the other methods described below to handle the result more safely.
+Like `get`, this will throw an exception if the index doesn't exist. Use the Simple Path `\` operator and `validate` or `asOpt` (described below) if you expect that they key may not be present.
 
 ## Reading and writing objects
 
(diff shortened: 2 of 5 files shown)


lihaoyi:
Test failure looks like a flake

gmethvin:
@lihaoyi For the examples we probably want to update the test here instead of creating a new test: https://github.com/playframework/play-json/blob/master/docs/manual/working/scalaGuide/main/json/code/ScalaJsonSpec.scala#L14

The compiled docs are here: https://www.playframework.com/documentation/2.6.x/ScalaJson, and the markdown source is here: https://github.com/playframework/play-json/blob/master/docs/manual/working/scalaGuide/main/json/ScalaJson.md

I'm not sure what to do for the README. I wanted to have a simple getting started guide in this repo, with the more complex examples in the compiled Play documentation.

gmethvin:
The idea is that `JsLookup` is a "safe" lookup. So it should return a `JsDefined`/`JsUndefined` depending on whether the key exists. This is effectively the way it's always been in play-json, though `JsUndefined` used to be included in the `JsValue` type.

It seems like the `JsLookup.apply` should be deprecated, since it's easily confused for `JsArray.apply`, and we suggest using `\` instead.

lihaoyi:
Thanks for the pointer to the tests in the `doc/` folder, I've updated them and got rid of the redundant tests I added.

I'm not sure what `JsArray#apply` method you are referring to? I don't see any in the codebase.

From a API-design perspective, I think that it makes sense to have an "unsafe" lookup easily available: we already have `\` to be the "safe" lookup, and my experience there are often use-cases where you just want to pull out data you know exists, and if it doesn't exist you want your program to crash with a (hopefully detailed) error message.

Perhaps that occurs more in small scripts than in large web applications, but I think it would be nice to be able to use play-json in small scripts rather than have to pull in a totally separate library like uPickle, especially if the necessary changes are so simple/superficial.

From an understandability point of view, I think people would not be confused by the difference between `\` and `.apply`: `.apply` exists in plain Scala `Map`s, `Array`s, `Vector`s, etc. and in all cases does the same thing this diff would make it do here: it looks up some key/index, and throws if it's not found. I think adding similar functionality to play-json would go a long way to help make the library easy to use in less-strict contexts like scripts, scrapers, dev-ops automation or the interactive REPL where I would just be calling `.get` everywhere anyway.

I understand that this is a non-trivial change in the API, so I don't expect you to merge it instantly, but I think it is a relatively-simple change that shouldn't confuse people and will make play-json much more convenient to use directly in less-strict contexts

gmethvin:
> I'm not sure what JsArray#apply method you are referring to? I don't see any in the codebase.

Sorry, for some reason I thought there was an unsafe `JsArray#apply` method already.

I'm okay with this change as long as we document the fact that both `apply` methods are "unsafe". When people upgrade, existing code will fail to compile, so we will just want to add something to the Play 2.6 migration guide to help people upgrading.

docs/manual/working/scalaGuide/main/json/ScalaJson.md
```diff
@@ -101,22 +101,26 @@ You can traverse a `JsValue` structure and extract specific values. The syntax a
 
 ### Simple path `\`
 
-Applying the `\` operator to a `JsValue` will return the property corresponding to the field argument, supposing this is a `JsObject`.
+Applying the `\` operator to a `JsValue` will return the property corresponding to the field argument in a `JsObject`, or the item at that index in a `JsArray`
 
 @[traverse-simple-path](code/ScalaJsonSpec.scala)
 
+In this example we are using `.get`, which retrieves the value out of the resulting `JsResult` and throws an exception if it was not found. You can also use the `JsResult` directly to exceptions, or use the Direct lookup `.apply` method below if you want a missing key or index to throw.
```
gmethvin: Here's a suggestion to clarify (note that `JsResult` is the result of parsing JSON, while `JsLookupResult` is the result of a lookup):

The `\` operator returns a `JsLookupResult`, which is either `JsDefined` or `JsUndefined`. You can chain multiple `\` operators, and the result will be `JsUndefined` if any intermediate value cannot be found. Calling `get` on a `JsLookupResult` attempts to get the value if it is defined and throws an exception if it does not exist.

You can also use the `apply` method to get a field in an object or index in an array. Like `get`, this method will throw an exception if the value does not exist.

lihaoyi:
I have updated the docs as you suggested.

I assume the play 2.6 migration guide would be in a separate repo somewhere else?

docs/manual/working/scalaGuide/main/json/ScalaJson.md
```diff
@@ -101,22 +101,29 @@ You can traverse a `JsValue` structure and extract specific values. The syntax a
...
 
 Applying the `\\` operator will do a lookup for the field in the current object and all descendants.
 
 @[traverse-recursive-path](code/ScalaJsonSpec.scala)
 
-### Index lookup (for JsArrays)
+### Direct lookup
 
-You can retrieve a value in a `JsArray` using an apply operator with the index number.
+You can retrieve a value in a `JsArray` or `JsObject` using an `.apply` operator, which is identical to the Simple path `\` operator except it returns the value directly (rather than wrapping it in a `JsResult`) throws an exception if the index or key is not found:
```
gmethvin: Should be:

> which is identical to the *s*imple path `\` operator except it returns the value directly (rather than wrapping it in a *`JsLookupResult`*)* and* throws an exception if the index or key is not found

(changes italicised)

lihaoyi:
Oops, added back the missing `and`

gmethvin:
Yeah, the migration guide is here: https://github.com/playframework/playframework/blob/master/documentation/manual/releases/release26/migration26/Migration26.md

I created an issue: playframework/playframework#7348. If you want you can resolve it, otherwise I'll get to it when I make another sweep through the migration docs in the next few weeks.

docs/manual/working/scalaGuide/main/json/ScalaJson.md
```diff
@@ -101,22 +101,29 @@ You can traverse a `JsValue` structure and extract specific values. The syntax a
...
 
 Applying the `\\` operator will do a lookup for the field in the current object and all descendants.
 
 @[traverse-recursive-path](code/ScalaJsonSpec.scala)
 
-### Index lookup (for JsArrays)
+### Direct lookup
 
-You can retrieve a value in a `JsArray` using an apply operator with the index number.
+You can retrieve a value in a `JsArray` or `JsObject` using an `.apply` operator, which is identical to the Simple path `\` operator except it returns the value directly (rather than wrapping it in a `JsResult`) and throws an exception if the index or key is not found:
```
gmethvin: Sorry should have been more clear. It's JsLookupResult not JsResult.

lihaoyi:
Changed `JsResult` -> `JsLookupResult`

lihaoyi:
Thank @gmethvin!

PR #67, opened 2017-05-12
Make `JsArray` take an `IndexedSeq` instead of a `Seq`

lihaoyi:
# Pull Request Checklist

## Fixes

Didn't find any open issue

## Purpose

This should make performance and memory characteristis much
more predictable. Before this diff, the following two lines of
"equivalent" code result in hugely different memory characteristics:

```scala
// 240mb
val json = Json.parse("[" + Array.fill(10000000)("true").mkString(",") + "]")

// 40mb
val json = Json.arr(Array.fill(10000000)(true: Json.JsValueWrapper):_*)
```

And performance characteristics:

```scala
val json = Json.parse("[" + Array.fill(10000)("true").mkString(",") + "]")
val start = System.currentTimeMillis()
var i = 0
while(i < 1000000){ json(9999); i += 1}
println(System.currentTimeMillis() - start) // 69405
```
```scala
val json = Json.arr(Array.fill(10000)(true: Json.JsValueWrapper):_*)
val json = Json.parse("[" + Array.fill(10000)("true").mkString(",") + "]")
val start = System.currentTimeMillis()
var i = 0
while(i < 1000000){ json(9999); i += 1}
println(System.currentTimeMillis() - start) // 172
```

While this is not a problem if we only expect to ever iterate over and
transform a `JsArray`, a user might expect to keep `JsArray`s in memory,
and likely expects to be able to index into them in constant time. After
all, that is possible in most JSON implementations, and the name "JsARRAY"
also hints that you should be able to look things up quickly.

With this diff, both cases behave identically, with low ~40mb memory usage
in the memory benchmark and fast 100-200ms performance in the lookup
benchmark.

We implement some of the `IndexedSeq`s using `Array`s and others using
`mutable.ArrayBuffer`s. There's still going to be some variation in
runtime characteristics, but I think it's probably
fine since it's a constant-time difference in terms of memory usage and
runtime performance. If we really cared, we could make sure to transform
everything into wrapped `Array`s before storing them.

The only casualties I found were the `Stream -> JsArray` tests:

```scala
    "validate JsArray of stream to List" in {
      JsArray(Stream("alpha", "beta", "delta") map JsString.apply).validate[List[String]] mustEqual (JsSuccess(List("alpha", "beta", "delta")))
    }

    "invalidate JsArray of stream to List with wrong type conversion" in {
      JsArray(Stream(JsNumber(1), JsString("beta"), JsString("delta"), JsNumber(4), JsString("five"))).validate[List[Int]] mustEqual (
        JsError(Seq(
          JsPath(1) -> Seq(JsonValidationError("error.expected.jsnumber")),
          JsPath(2) -> Seq(JsonValidationError("error.expected.jsnumber")),
          JsPath(4) -> Seq(JsonValidationError("error.expected.jsnumber"))
        ))
      )

      JsArray(Stream(JsString("alpha"), JsNumber(5), JsBoolean(true))).validate[List[Int]] mustEqual (
        JsError(Seq(
          JsPath(0) -> Seq(JsonValidationError("error.expected.jsnumber")),
          JsPath(2) -> Seq(JsonValidationError("error.expected.jsnumber"))
        ))
      )
    }
```

I personally think this is not a problem, as `JsArray` constructor are
not documented as public APIs on the github page (though they seem to be exposed [here](https://www.playframework.com/documentation/2.5.x/ScalaJson#Using-class-construction)), and users seem to be pushed towards using
`Json.arr` and friends. Nonetheless, if we think this is problematic, we
could add an overloaded version of `JsArray.apply` to maintain source
compatibility. Unfortunately, binary compatibility is probably a lost cause
with this kind of diff.

Like https://github.com/playframework/play-json/pull/66, this is submitted without any prior discussion or bug reports. Nonetheless, It seems like it's the right thing to do, and it was pretty easy to make these changes to show off even if it doesn't end up landing. Let me know what you think!

### DIFF
--- a/docs/manual/working/scalaGuide/main/json/code/ScalaJsonSpec.scala
+++ b/docs/manual/working/scalaGuide/main/json/code/ScalaJsonSpec.scala
@@ -56,7 +56,7 @@ class ScalaJsonSpec extends Specification {
       val json: JsValue = JsObject(Seq(
         "name" -> JsString("Watership Down"),
         "location" -> JsObject(Seq("lat" -> JsNumber(51.235685), "long" -> JsNumber(-1.309197))),
-        "residents" -> JsArray(Seq(
+        "residents" -> JsArray(IndexedSeq(
           JsObject(Seq(
             "name" -> JsString("Fiver"),
             "age" -> JsNumber(4),
--- a/play-json/js/src/main/scala/StaticBinding.scala
+++ b/play-json/js/src/main/scala/StaticBinding.scala
@@ -108,7 +108,7 @@ object StaticBinding {
     case l: Long => JsNumber(l)
     case true => JsTrue
     case false => JsFalse
-    case a: js.Array[_] => JsArray(a.map(anyToJsValue).toSeq)
+    case a: js.Array[_] => JsArray(a.map(anyToJsValue).toArray[JsValue])
 
     case o: js.Object => {
       JsObject((for {
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -17,7 +17,8 @@ import com.fasterxml.jackson.databind.ser.Serializers
 import play.api.libs.json._
 
 import scala.annotation.{ switch, tailrec }
-import scala.collection.mutable.ListBuffer
+import scala.collection.mutable
+import scala.collection.mutable.{ ArrayBuffer, ListBuffer }
 
 /**
  * The Play JSON module for Jackson.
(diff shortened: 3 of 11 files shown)


gmethvin:
I'm ok with this change, but could we also provide an alternate `apply` method that does the conversion from an arbitrary `Seq`? That would cover most of the backwards compatibility issues.

I've seen `JsArray.apply` is used in Play code a lot, so I'm not comfortable removing that method completely without deprecation.

gmethvin:
For comparison see what we did with `JsObject`. We're actually using a `mutable.LinkedHashMap` for performance reasons there (because some people want predictable field order, even though it's not in the JSON spec), but we also have a constructor that accepts a `Seq[(String, JsValue)]`

lihaoyi:
I've added the overload to `JsArray.apply(values: Seq[JsValue])`; the old tests pass again now

gmethvin:
Thanks @lihaoyi

PR #74, opened 2017-06-27
Fix warnings when checking apply methods

### LINKED ISSUE #73: Case class with several similar `apply` functions
I tried to upgrade from "2.5.14" to "2.6.0" and stumbled on interesting issue when case class has several similar `apply` functions.

```
package xxx

import play.api.libs.json.{Json, OFormat}

case class XXX(xxx: String, age: Int)

object XXX {
  implicit val XxxFormat: OFormat[XXX] = Json.format[XXX]
}
```

Compiles fine, until one adds second `apply` function like:
```
  def apply(xxx: String, age: Option[Int]): XXX = XXX(xxx, age getOrElse 21)
```

When above code is added in companion object, I am getting warning: 
```
Warning:(8, 53) Type parameters are not matching: Option[Int] != Int
  implicit val XxxFormat: OFormat[XXX] = Json.format[XXX]
```

In our project we have many such places which are working fine with "2.5.14", and we have enabled `"-Xfatal-warnings"` in scalac options - this option promotes this warning to compilation error.

What is suggested way to migrate such code to 2.6.0 version? 
Maybe I am doing something wrong here?

gmethvin:
Since the `conforms` method can be called for all `apply` methods with the correct number of parameters, it does not make sense to warn in this method if the parameters don't match, since it may be the case that there's another apply method that matches. Instead I changed the `warning` to `info`.

Fixes #73

### DIFF
--- a/play-json/shared/src/main/scala/JsMacroImpl.scala
+++ b/play-json/shared/src/main/scala/JsMacroImpl.scala
@@ -240,9 +240,10 @@ import scala.reflect.macros.blackbox
           case Some((TypeRef(NoPrefix, a, _),
             TypeRef(NoPrefix, b, _))) => { // for generic parameter
             if (a.fullName != b.fullName) {
-              c.warning(
+              c.info(
                 c.enclosingPosition,
-                s"Type symbols are not compatible: $a != $b"
+                s"Type symbols are not compatible: $a != $b",
+                force = false
               )
 
               false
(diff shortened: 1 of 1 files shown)


mr-git:
I am not sure that this is the best possible fix.
IMHO it would be nicer, if it would be possible to do check against all `apply` functions and do print error only when there is not a single match.

cchantep:
@mr-git the global check is already done : https://github.com/gmethvin/play-json/blob/f2e73ba2677132753a9c6ac0ff3ea01131e2d7ea/play-json/shared/src/main/scala/JsMacroImpl.scala#L596

mr-git:
Good!
Then I wonder, why one needs those extra info lines in log? They do not carry any meaningful or useful information and could cause misunderstanding.

cchantep:
They carry meaningful information, in case no `apply` is found to check which one were checked to lead to such situation.

PR #84, opened 2017-07-20
Added strict properties feature

jroper:
This adds a new feature flag to the macro opts that allows enabling strict properties - that is, the produced JSON reads/format will fail if any unknown properties are encountered. This is useful particularly in testing if you have a JSON structure pulled from a system that you're consuming data from, and you want to verify that your case classes correctly match it, and haven't misnamed any optional properties.

### DIFF
--- a/play-json/shared/src/main/scala/JsMacroImpl.scala
+++ b/play-json/shared/src/main/scala/JsMacroImpl.scala
@@ -3,6 +3,8 @@
  */
 package play.api.libs.json
 
+import play.api.libs.json.Json.StrictProperties
+
 import scala.language.experimental.macros
 import scala.language.higherKinds
 import scala.reflect.macros.blackbox
@@ -617,14 +619,19 @@ import scala.reflect.macros.blackbox
           }.toMap
         }
 
+      // Determine the JSON property name for a given symbol name
+      def configNaming(name: Name): c.Expr[String] = {
+        c.Expr[String](
+          q"$cfgName.naming(${name.decodedName.toString})"
+        )
+      }
+
       val resolvedImplicits = utility.implicits(resolver)
       val canBuild = resolvedImplicits.map {
         case (name, Implicit(pt, impl, _, _)) =>
           // Equivalent to __ \ "name", but uses a naming scheme
           // of (String) => (String) to find the correct "name"
-          val cn = c.Expr[String](
-            q"$cfgName.naming(${name.decodedName.toString})"
-          )
+          val cn = configNaming(name)
           val jspathTree = q"$JsPath \ $cn"
           val isOption = pt.typeConstructor <:< optTpeCtor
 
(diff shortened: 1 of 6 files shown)


play-json/shared/src/main/scala/JsMacroImpl.scala
```diff
@@ -667,33 +674,49 @@ import scala.reflect.macros.blackbox
...
+        } else {
+          theReads
+        }
+      }
 
       val canBuildCall = methodName match {
-        case "read" => {
+        case "read" =>
+          val underlying = TermName(c.freshName("underlying"))
+          val obj = TermName(c.freshName("obj"))
```
jroper: Just a small hygiene change, macros shouldn't just define arbitrary terms, they should synthesis them with freshName.

play-json/shared/src/main/scala/Json.scala
```diff
@@ -360,7 +360,7 @@ object Json extends JsonFacade {
      *   Json.using[Json.MacroOptions].writes[User]
      * }}}
      */
-    def writes[A]: OWrites[A] = macro JsMacroImpl.writesImpl[A, MacroOptions]
+    def writes[A]: OWrites[A] = macro JsMacroImpl.writesImpl[A, Opts]
```
jroper: I'm pretty sure this was a bug, I just happened to notice it.

play-json/shared/src/main/scala/Reads.scala
```diff
@@ -192,6 +192,22 @@ trait DefaultReads extends LowPriorityDefaultReads {
...
+      val unknown = jsObject.keys -- knownProperties
+      if (unknown.isEmpty) {
+        JsSuccess(jsObject)
+      } else {
+        JsError(unknown.map { prop =>
+          __ \ prop -> Seq(JsonValidationError("error.unexpected.property"))
+        }.toSeq)
+      }
+    }
+  }
```
jroper: Defined here as a reusable reads so anyone can use it.

play-json/shared/src/main/scala/JsMacroImpl.scala
```diff
@@ -617,14 +619,19 @@ import scala.reflect.macros.blackbox
           }.toMap
         }
 
+      // Determine the JSON property name for a given symbol name
+      def configNaming(name: Name): c.Expr[String] = {
+        c.Expr[String](
+          q"$cfgName.naming(${name.decodedName.toString})"
+        )
+      }
```
jroper: Extracted for reuse.

play-json/shared/src/main/scala/Json.scala
```diff
@@ -432,4 +432,19 @@ object Json extends JsonFacade {
    * }}}
    */
   type WithDefaultValues = MacroOptions with DefaultValues
+
+  /**
+   * Flag to indicate that the macros should generate reads
+   * that fail on unknown properties.
+   */
```
cchantep: Example? `MacroOptions with ...`

play-json/shared/src/main/scala/Json.scala
cchantep: Good catch (not detected by tests as write is unaffected by options for now)

play-json/shared/src/main/scala/Reads.scala
```diff
@@ -129,6 +129,22 @@ object Reads extends ConstraintReads with PathReads with DefaultReads with Gener
   }
 
   implicit val JsArrayReducer = Reducer[JsValue, JsArray](js => JsArray(Array(js)))
+
+  /**
+   * A reads that fails if any of the objects properties are not present in the passed in known properties list.
+   */
+  def failOnUnknownProperties(knownProperties: immutable.Seq[String]): Reads[JsObject] = Reads { jsValue =>
```
cchantep: - Directly `JsObjectReads.flatMap`?
- Does the order of properties is/need to be checked? If not maybe passing them as `Set` would make sense

play-json/shared/src/main/scala/Reads.scala
```diff
@@ -129,6 +129,22 @@ object Reads extends ConstraintReads with PathReads with DefaultReads with Gener
...
+    JsObjectReads.reads(jsValue).flatMap { jsObject =>
+      val unknown = jsObject.keys -- knownProperties
+      if (unknown.isEmpty) {
+        JsSuccess(jsObject)
+      } else {
+        JsError(unknown.map { prop =>
+          __ \ prop -> Seq(JsonValidationError("error.unexpected.property"))
```
cchantep: `JsonValidationError(..., prop)` (so property name can be substituted in the message)

jroper: I don't think that's needed - the property name is available as the path (that's why it's `__ \ prop`) putting it in the message would just be duplicate information.

cchantep: right

gmethvin: Actually it might be useful to substitute the list of allowed properties in the message.

play-json/shared/src/main/scala/Reads.scala
gmethvin: Yeah I think `Set` makes more sense. Or even `String => Boolean`. Though if we're going to print the allowed properties in the error message being able to control the ordering might be nice.

cchantep: `failOnUnknownProperty` for naming (if fails on the first one) ?

SethTisue:
closing for inactivity. happy to reopen if activity resumes. (or, someone could submit a new PR)

zhangchuanben:
That feature is great, can i use `strict-properties` feature now ? @jroper

PR #85, opened 2017-07-20
Improved configuration passing

jroper:
The current implementation of passing configuration and macro options had a major problem. You could only pass a naming strategy by making an implicit JsonConfiguration variable available, while you could only pass a macro option using Json.using or Json.configured. If you used the wrong one for the wrong approach, then it didn't work.  So something like this:

```scala
implicit val config = JsonConfiguration[WithDefaultValues](SnakeCase)
```

Would use snake case, but not default values, while:

```scala
Json.configured(JsonConfiguration[WithDefaultValues](SnakeCase))
```

Would turn on default values, but not use snake case. Very confusing for end users.

This fixes all that, now either method can be used to pass either configuration or macro options.

### DIFF
--- a/play-json/shared/src/main/scala/JsMacroImpl.scala
+++ b/play-json/shared/src/main/scala/JsMacroImpl.scala
@@ -26,11 +26,42 @@ import scala.reflect.macros.blackbox
     }
   }
 
-  def readsImpl[A: c.WeakTypeTag, O <: Json.MacroOptions: c.WeakTypeTag]: c.Expr[Reads[A]] = macroImpl[A, Reads, Reads, O]("read", "map", reads = true, writes = false)
+  def withOptionsReadsImpl[A: c.WeakTypeTag]: c.Expr[Reads[A]] = macroImpl[A, Reads, Reads](
+    withOptionsConfig, "read", "map", reads = true, writes = false)
 
-  def writesImpl[A: c.WeakTypeTag, O <: Json.MacroOptions: c.WeakTypeTag]: c.Expr[OWrites[A]] = macroImpl[A, OWrites, Writes, O]("write", "contramap", reads = false, writes = true)
+  def withOptionsWritesImpl[A: c.WeakTypeTag]: c.Expr[OWrites[A]] = macroImpl[A, OWrites, Writes](
+    withOptionsConfig, "write", "contramap", reads = false, writes = true)
 
-  def formatImpl[A: c.WeakTypeTag, O <: Json.MacroOptions: c.WeakTypeTag]: c.Expr[OFormat[A]] = macroImpl[A, OFormat, Format, O]("format", "inmap", reads = true, writes = true)
+  def withOptionsFormatImpl[A: c.WeakTypeTag]: c.Expr[OFormat[A]] = macroImpl[A, OFormat, Format](
+    withOptionsConfig, "format", "inmap", reads = true, writes = true)
+
+  def implicitConfigReadsImpl[A: c.WeakTypeTag]: c.Expr[Reads[A]] = macroImpl[A, Reads, Reads](
+    implicitOptionsConfig, "read", "map", reads = true, writes = false)
+
+  def implicitConfigWritesImpl[A: c.WeakTypeTag]: c.Expr[OWrites[A]] = macroImpl[A, OWrites, Writes](
+    implicitOptionsConfig, "write", "contramap", reads = false, writes = true)
+
+  def implicitConfigFormatImpl[A: c.WeakTypeTag]: c.Expr[OFormat[A]] = macroImpl[A, OFormat, Format](
+    implicitOptionsConfig, "format", "inmap", reads = true, writes = true)
+
+  // because binary compatibility...
+  @deprecated("Use implicitConfigReadsImpl or withOptionsReadsImpl", "2.6.6")
+  protected def readsImpl[A: c.WeakTypeTag, O: c.WeakTypeTag]: c.Expr[Reads[A]] = macroImpl[A, Reads, Reads](
+    implicitOptionsConfig, "read", "map", reads = true, writes = false)
+  @deprecated("Use implicitConfigWritesImpl or withOptionsWritesImpl", "2.6.6")
+  protected def writesImpl[A: c.WeakTypeTag, O: c.WeakTypeTag]: c.Expr[OWrites[A]] = macroImpl[A, OWrites, Writes](
+    implicitOptionsConfig, "write", "contramap", reads = false, writes = true)
+  @deprecated("Use implicitConfigFormatImpl or withOptionsFormatImpl", "2.6.6")
+  protected def formatImpl[A: c.WeakTypeTag, O: c.WeakTypeTag]: c.Expr[OFormat[A]] = macroImpl[A, OFormat, Format](
+    implicitOptionsConfig, "format", "inmap", reads = true, writes = true)
+
+  private def withOptionsConfig: c.Expr[JsonConfiguration] = {
+    c.Expr[JsonConfiguration](c.typecheck(q"${c.prefix}.config"))
+  }
+
+  private def implicitOptionsConfig: c.Expr[JsonConfiguration] = {
+    c.Expr[JsonConfiguration](c.inferImplicitValue(c.typeOf[JsonConfiguration]))
+  }
 
   /**
    * Generic implementation of the macro.
@@ -40,7 +71,7 @@ import scala.reflect.macros.blackbox
    * functions, for example, if the apply, unapply, or both should be passed to the functional builder apply
    * method.
    *
-   * @param c The context
+   * @param config The configuration tree
    * @param methodName The name of the method on JsPath that gets called, ie, read/write/format
    * @param mapLikeMethod The method that's used to map the type of thing being built, used in case there is
    * only one field in the case class.
@@ -50,7 +81,8 @@ import scala.reflect.macros.blackbox
    * @param matag The class of the reads/writes/format.
    * @param natag The class of the reads/writes/format.
    */
-  private def macroImpl[A, M[_], N[_], O <: Json.MacroOptions](methodName: String, mapLikeMethod: String, reads: Boolean, writes: Boolean)(implicit atag: c.WeakTypeTag[A], matag: c.WeakTypeTag[M[A]], natag: c.WeakTypeTag[N[A]], otag: c.WeakTypeTag[O]): c.Expr[M[A]] = {
+  private def macroImpl[A, M[_], N[_]](config: c.Expr[JsonConfiguration], methodName: String, mapLikeMethod: String,
+    reads: Boolean, writes: Boolean)(implicit atag: c.WeakTypeTag[A], matag: c.WeakTypeTag[M[A]], natag: c.WeakTypeTag[N[A]]): c.Expr[M[A]] = {
 
     def debug(msg: => String): Unit = {
       if (debugEnabled) {
@@ -84,7 +116,7 @@ import scala.reflect.macros.blackbox
     val forwardName = TermName(c.freshName("forward"))
 
     // MacroOptions
-    val options = c.weakTypeOf[O]
+    val options = config.actualType.member(TypeName("Opts")).asType.toTypeIn(config.actualType)
     def hasOption[Flag: c.TypeTag]: Boolean = options <:< typeOf[Flag]
 
     /* Utility for implicit resolution - can hardly be moved outside
@@ -604,7 +636,6 @@ import scala.reflect.macros.blackbox
       // ---
 
       // combines all reads into CanBuildX
-      val cfgName = TermName(c.freshName("config"))
       val resolver = new ImplicitResolver({
         import utility.boundTypes
 
@@ -626,7 +657,7 @@ import scala.reflect.macros.blackbox
           // Equivalent to __ \ "name", but uses a naming scheme
           // of (String) => (String) to find the correct "name"
           val cn = c.Expr[String](
-            q"$cfgName.naming(${name.decodedName.toString})"
+            q"$config.naming(${name.decodedName.toString})"
           )
           val jspathTree = q"$JsPath \ $cn"
           val isOption = pt.typeConstructor <:< optTpeCtor
@@ -707,8 +738,6 @@ import scala.reflect.macros.blackbox
           q"""
         $syntaxImport
 
-        val $cfgName = implicitly[$json.JsonConfiguration]
-
         $canBuildCall
         """
         } else {
@@ -730,7 +759,7 @@ import scala.reflect.macros.blackbox
           val generated = TypeName(c.freshName("Generated"))
 
           q"""
-        final class $generated()(implicit $cfgName: $json.JsonConfiguration) {
+        final class $generated() {
           // wrap there for self reference
 
           $syntaxImport
--- a/play-json/shared/src/main/scala/Json.scala
+++ b/play-json/shared/src/main/scala/Json.scala
@@ -6,6 +6,8 @@ package play.api.libs.json
 
 import java.io.InputStream
 
+import play.api.libs.json.Json.MacroOptions
+
 /**
  * @define jsonParam @param json the JsValue to convert
  * @define returnStringRepr A String with the json representation
@@ -244,7 +246,7 @@ object Json extends JsonFacade {
    * )(User)
    * }}}
    */
-  def reads[A]: Reads[A] = macro JsMacroImpl.readsImpl[A, MacroOptions]
+  def reads[A]: Reads[A] = macro JsMacroImpl.implicitConfigReadsImpl[A]
 
   /**
    * Creates a `OWrites[T]` by resolving, at compile-time,
@@ -268,7 +270,7 @@ object Json extends JsonFacade {
    *   )(unlift(User.unapply))
    * }}}
    */
-  def writes[A]: OWrites[A] = macro JsMacroImpl.writesImpl[A, MacroOptions]
+  def writes[A]: OWrites[A] = macro JsMacroImpl.implicitConfigWritesImpl[A]
 
   /**
    * Creates a `OFormat[T]` by resolving, at compile-time,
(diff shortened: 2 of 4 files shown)


play-json/shared/src/main/scala/JsMacroImpl.scala
```diff
@@ -25,11 +25,39 @@ import scala.reflect.macros.blackbox
...
+  def withOptionsReadsImpl[A: c.WeakTypeTag]: c.Expr[Reads[A]] = macroImpl[A, Reads, Reads](
+    withOptionsConfig, "read", "map", reads = true, writes = false)
 
-  def writesImpl[A: c.WeakTypeTag, O <: Json.MacroOptions: c.WeakTypeTag]: c.Expr[OWrites[A]] = macroImpl[A, OWrites, Writes, O]("write", "contramap", reads = false, writes = true)
+  def withOptionsWritesImpl[A: c.WeakTypeTag]: c.Expr[OWrites[A]] = macroImpl[A, OWrites, Writes](
+    withOptionsConfig, "write", "contramap", reads = false, writes = true)
 
-  def formatImpl[A: c.WeakTypeTag, O <: Json.MacroOptions: c.WeakTypeTag]: c.Expr[OFormat[A]] = macroImpl[A, OFormat, Format, O]("format", "inmap", reads = true, writes = true)
+  def withOptionsFormatImpl[A: c.WeakTypeTag]: c.Expr[OFormat[A]] = macroImpl[A, OFormat, Format](
+    withOptionsConfig, "format", "inmap", reads = true, writes = true)
```
jroper: These new variants are for invoking from `WithOptions.reads/writes/format`. They assume that the context they are invoked from as a `config: JsonConfiguration.Aux[Opts]` property on them.

play-json/shared/src/main/scala/JsMacroImpl.scala
```diff
@@ -25,11 +25,39 @@ import scala.reflect.macros.blackbox
...
+
+  def readsImpl[A: c.WeakTypeTag]: c.Expr[Reads[A]] = macroImpl[A, Reads, Reads](
+    implicitOptionsConfig, "read", "map", reads = true, writes = false)
+
+  def writesImpl[A: c.WeakTypeTag]: c.Expr[OWrites[A]] = macroImpl[A, OWrites, Writes](
+    implicitOptionsConfig, "write", "contramap", reads = false, writes = true)
+
+  def formatImpl[A: c.WeakTypeTag]: c.Expr[OFormat[A]] = macroImpl[A, OFormat, Format](
+    implicitOptionsConfig, "format", "inmap", reads = true, writes = true)
```
jroper: These are the old variants, used by `Json.reads/writes/format`, they discover the configuration via an implicit value.

play-json/shared/src/main/scala/JsMacroImpl.scala
```diff
@@ -25,11 +25,39 @@ import scala.reflect.macros.blackbox
...
+
+  // because binary compatibility...
+  protected def readsImpl[A: c.WeakTypeTag, O: c.WeakTypeTag]: c.Expr[Reads[A]] = macroImpl[A, Reads, Reads](
+    implicitOptionsConfig, "read", "map", reads = true, writes = false)
+  protected def writesImpl[A: c.WeakTypeTag, O: c.WeakTypeTag]: c.Expr[OWrites[A]] = macroImpl[A, OWrites, Writes](
+    implicitOptionsConfig, "write", "contramap", reads = false, writes = true)
+  protected def formatImpl[A: c.WeakTypeTag, O: c.WeakTypeTag]: c.Expr[OFormat[A]] = macroImpl[A, OFormat, Format](
+    implicitOptionsConfig, "format", "inmap", reads = true, writes = true)
```
jroper: Because the old methods are public, changing their signature breaks binary compatibility. So I put the old methods here, but made them protected, and Scala protected methods compile to public Java methods.

play-json/shared/src/main/scala/JsMacroImpl.scala
```diff
@@ -25,11 +25,39 @@ import scala.reflect.macros.blackbox
...
+
+  private def withOptionsConfig: Tree = {
+    c.typecheck(q"${c.prefix}.config")
+  }
```
jroper: Read the config from where the macro is defined, ie, from `WithOptions`.

play-json/shared/src/main/scala/JsMacroImpl.scala
```diff
@@ -25,11 +25,39 @@ import scala.reflect.macros.blackbox
...
+
+  private def implicitOptionsConfig: Tree = {
+    c.inferImplicitValue(c.typeOf[JsonConfiguration])
+  }
```
jroper: Read the config implicitly.

play-json/shared/src/main/scala/JsMacroImpl.scala
```diff
@@ -77,7 +105,7 @@ import scala.reflect.macros.blackbox
     val forwardName = TermName(c.freshName("forward"))
 
     // MacroOptions
-    val options = c.weakTypeOf[O]
+    val options = config.tpe.member(TypeName("Opts")).asType.toTypeIn(config.tpe)
```
jroper: I hope this is the right way to resolve the type of a type member. @retronym?

play-json/shared/src/main/scala/Json.scala
```diff
@@ -301,7 +303,10 @@ object Json extends JsonFacade {
    * @define macroWarning If any missing implicit is discovered, compiler will break with corresponding error.
    * @define macroTypeParam @tparam A the type for which the handler must be materialized
    */
-  final class WithOptions[Opts <: MacroOptions]() extends JsonFacade {
+  final class WithOptions[Opts <: MacroOptions](val config: JsonConfiguration.Aux[Opts]) extends JsonFacade {
```
jroper: So now `WithOptions` keeps the `JsonConfiguration`, which allows the macros to read it.

play-json/shared/src/main/scala/Json.scala
```diff
@@ -301,7 +303,10 @@ object Json extends JsonFacade {
...
+
+    private[json] def this() = this(JsonConfiguration.default)
```
jroper: Binary compatibility. `private[json]` is public in byte code.

play-json/shared/src/main/scala/Json.scala
```diff
@@ -393,15 +398,15 @@ object Json extends JsonFacade {
    * val r: Reads[Foo] = Json.configured.reads[Foo]
    * }}}
    */
-  def configured[C <: JsonConfiguration.Aux[_ <: MacroOptions]](implicit config: C) = new WithOptions[config.Opts]()
+  def configured[Opts <: MacroOptions](implicit config: JsonConfiguration.Aux[Opts]) = new WithOptions[Opts](config)
```
jroper: Not sure what the go with the old type was, but what we really needed was to return a `WithOptions` that has the same `MacroOptions` type as the passed in configuration, this does that.

play-json/shared/src/main/scala/Json.scala
```diff
@@ -414,6 +419,29 @@ object Json extends JsonFacade {
...
+
+  object DefaultMacroOptions extends NotDefaultMacroOptions {
+
+    /**
+     * This wil be the default that's passed, causing Json.MacroOptions to be the default options.
+     */
+    implicit object defaultMacroOptions extends DefaultMacroOptions[Json.MacroOptions]
+  }
+
+  /**
```
jroper: The comments pretty much say it all, but just to clarify, consider below, the:

```scala
implicit def default[Opts <: Json.MacroOptions]: JsonConfiguration.Aux[Json.MacroOptions]`
```

on `JsonConfiguration`. If I just invoke that without supplying a type (as gets down when it's invoke implicitly), what type do I get back?  The answer is, `JsonConfiguration.Aux[Nothing]`. And what does that mean? I'm not 100% sure, but I think it means that every option is enabled, since `Nothing` is a subclass of everything. That's a pretty serious bug. Ideally, I'd like to change the signature to:

```scala
implicit def default[Opts <: Json.MacroOptions = Json.MacroOptions]: JsonConfiguration.Aux[Json.MacroOptions]`
```

ie, supply a default type parameter, but Scala doesn't support that feature. So instead we supply the default via an implicit. This solution comes courtesy of @travisbrown (thanks!):

https://stackoverflow.com/questions/25127140/how-can-parameters-settings-be-passed-to-a-scala-macro

play-json/shared/src/test/scala/JsonExtensionSpec.scala
```diff
@@ -712,5 +714,20 @@ class JsonExtensionSpec extends WordSpec with MustMatchers {
...
+      "allow supplying configuration via implicit config" in {
+        implicit val config = JsonConfiguration[Json.WithDefaultValues](naming = SnakeCase)
+        json.as(Json.reads[WithDefaultSnake]) mustEqual data
+      }
+
+      "allow supplying configuration via WithOptions" in {
+        val config = JsonConfiguration[Json.WithDefaultValues](naming = SnakeCase)
+        json.as(Json.configured(config).reads[WithDefaultSnake]) mustEqual data
+      }
+    }
```
jroper: And here's the results, we can see that both the naming, and the macro options, can be passed either via an implicit config, or via the `WithOptions` implementation of `JsonFacade`.

play-json/shared/src/main/scala/Json.scala
```diff
@@ -414,6 +419,29 @@ object Json extends JsonFacade {
...
   /**
+   * Defines the default macro options.
+   *
+   * Since you can't have defaults for a type parameter (unless it's a contravariant type, it will default to
+   * Nothing), we supply the default via an implicit parameter.
+   */
+  trait DefaultMacroOptions[O <: Json.MacroOptions]
+
+  /**
+   * Low priority implicit used when Json.MacroOptions is not passed.
```
cchantep: ? When there is no explicit `MacroOptions` passed I would expect the `DefaultMacroOptions` to be used

play-json/shared/src/main/scala/Json.scala
```diff
@@ -414,6 +419,29 @@ object Json extends JsonFacade {
...
+   */
+  trait NotDefaultMacroOptions {
+    implicit def notDefault[O <: Json.MacroOptions]: DefaultMacroOptions[O] = new DefaultMacroOptions[O] {}
+  }
+
+  object DefaultMacroOptions extends NotDefaultMacroOptions {
```
cchantep: Sound weird when reading it ...

play-json/shared/src/main/scala/JsonConfiguration.scala
```diff
@@ -21,16 +21,25 @@ object JsonConfiguration {
...
+  protected def apply(
+    naming: JsonNaming
+  ): JsonConfiguration.Aux[Json.MacroOptions] = new Impl(naming)
+
   /**
-   * @tparam O the options for the JSON macros
    * @param naming the naming strategy
    */
-  def apply[O <: Json.MacroOptions](
+  def apply[Opts <: Json.MacroOptions: Json.DefaultMacroOptions](
```
cchantep: What's the goal there?

play-json/shared/src/main/scala/Json.scala
jroper: `JsonConfiguration.apply` now takes a `DefaultMacroOptions` type class, so an implicit for every type that you want to pass to it must be defined. The highest priority implicit is `DefaultMacroOptions[MacroOptions]`, this causes `MacroOptions` to be the default if nothing is explicitly specified. But if something that is not the default gets passed, then we also need to provide a `DefaultMacroOptions` implicit for it, that is defined on this trait, and it has to be in a trait to make it lower priority than the default.

play-json/shared/src/main/scala/JsonConfiguration.scala
jroper: See https://github.com/playframework/play-json/pull/85#discussion_r128500684

jroper: Previously, if you didn't supply a type, you'd end up with `JsonConfiguration.Aux[Nothing]`, and since `Nothing` is a bottom type, ie, it's a sub class of everything, that meant every option was enabled. So `WithDefaultValues` has been enabled by default - that's a bug. I think.

play-json/shared/src/main/scala/Json.scala
cchantep: Why does it takes a `DefaultMacroOptions` instead of `MacroOptions` ?

jroper: It does take a `MacroOptions` type, and a type class of `DefaultMacroOptions`. See the other comments, this is the simplest way to implement default types for type parameters. I want the type returned by `apply` to default to `MacroOptions` when no type is explicitly passed, not `Nothing`.

cchantep: I think the naming is bothering me. If I understand right, it's more an evidence typeclass, to avoid ambiguity around the `MacroOptions` type (could be manage with 2.12 `@implicitNotFound`?), rather that a typeclass defining by itself such options?

jroper: The names can be changed. It's an evidence typeclass that specifies what the default type of `Opt <: MacroOptions` should be when `Opt` is not explicitly supplied, working around the fact that the Scala compiler infers `Nothing`. But I'm not sure if there's a better name than `MacroOptionsDefault`, because it is all about the default `MacroOptions` type. We could change `NotDefaultMacroOptions` to `LowPriorityDefaultMacroOptions`, but the thing is that `NotDefaultMacroOptions` actually does supply the `DefaultMacroOptions` type class for `_ <: MacroOptions` types that are not the default, so it seems pretty descriptive to me.

cchantep: For me this new type doesn't represent the default options themselves, but the evidence that passed options conform with option type.

Would personally prefer a more neutral name (`Aux` ?)

jroper: We don't need evidence that the options conform to the option type, the type bounds already do that. If I invoke `JsonConfiguration[MyOptions]()`, the compiler is able to do all the necessary type checking and passing of types without this type class, without any evidence of anything. But, if I invoke `JsonConfiguration()`, then in the absence of any type class supplying evidence, `Nothing` is passed as the type. That's why this typeclass is needed, it supplies the evidence for what the default options should be when none is selected. If it weren't for that one use case, we could remove it.

cchantep: I think that's also what I mean: the typeclass is there to enforce the evidence that the `T <: MacroOptions` is not `Nothing`

gmethvin: I would call it `MacroOptionsDefault` instead of `DefaultMacroOptions`. `MacroOptionsDefault` suggests it's a new type of thing, as opposed to a default implementation of `MacroOptions`.

Instead of `NotDefaultMacroOptions` I think it's ok to call it `LowPriorityMacroOptionsDefaultImplicits`. The name is more verbose but the "LowPriority*" makes it more readable since it follows the standard naming convention for this type of trait.

@jroper It would also be useful to have an explanation like you provided in the comments here about how this works to provide an appropriate default value for `JsonConfiguration`.

play-json/shared/src/main/scala/Json.scala
gmethvin: Is there a reason we can't just keep this public for source compatibility as well?

gmethvin:
@cchantep @jroper I updated the naming of `DefaultMacroOptions` to `MacroOptions.Default` and clarified the scaladoc a little bit. Does that look ok?

play-json/shared/src/main/scala/JsMacroImpl.scala
cchantep: Add `@deprecated` ?

play-json/shared/src/main/scala/JsMacroImpl.scala
```diff
@@ -49,7 +77,7 @@ import scala.reflect.macros.blackbox
    * @param matag The class of the reads/writes/format.
    * @param natag The class of the reads/writes/format.
    */
-  private def macroImpl[A, M[_], N[_], O <: Json.MacroOptions](methodName: String, mapLikeMethod: String, reads: Boolean, writes: Boolean)(implicit atag: c.WeakTypeTag[A], matag: c.WeakTypeTag[M[A]], natag: c.WeakTypeTag[N[A]], otag: c.WeakTypeTag[O]): c.Expr[M[A]] = {
+  private def macroImpl[A, M[_], N[_]](config: Tree, methodName: String, mapLikeMethod: String, reads: Boolean, writes: Boolean)(implicit atag: c.WeakTypeTag[A], matag: c.WeakTypeTag[M[A]], natag: c.WeakTypeTag[N[A]]): c.Expr[M[A]] = {
```
cchantep: Note: Replace type parameter `O` with parameter `config`

cchantep: Passing config as `c.Expr[JsonConfiguration]` ?

play-json/shared/src/main/scala/Json.scala
cchantep: `@deprecated` ?

play-json/shared/src/main/scala/Json.scala
```diff
@@ -413,6 +418,31 @@ object Json extends JsonFacade {
...
   sealed trait MacroOptions
 
+  object MacroOptions {
+    /**
+     * Defines the default macro options if no type is supplied.
+     *
+     * Since you can't have defaults for a type parameter (unless it's a contravariant type, it will default to
+     * Nothing), we supply the default via an implicit parameter.
+     */
+    trait Default[O <: Json.MacroOptions]
```
cchantep: `Fallback` ?

gmethvin: I think the naming is going to be confusing no matter what, since the object itself is meaningless and just used for its type. Maybe I should add a link to https://stackoverflow.com/a/29415519?

play-json/shared/src/main/scala/JsMacroImpl.scala
cchantep: `c.Expr[JsonConfiguration](inferImplicitValue ...)` advantage would "just" be to keep the type up to the Macro impl

play-json/shared/src/main/scala/Json.scala
cchantep: No strong opinion. Fallback could also denote that it is something use only if no explicit options are defined

play-json/shared/src/main/scala/JsMacroImpl.scala
gmethvin: Ok I updated to use `c.Expr[JsonConfiguration]`

gmethvin:
2.6.x: 2998d39

PR #89, opened 2017-08-04
Close #88 - Improve nullable support

### LINKED ISSUE #88: wrong implicit resolution
```scala
import play.api.libs.json._

def dontSerialize[T](fallback: T): Format[T] = new Format[T]{
  def writes(v: T): JsValue = JsNull
  def reads(json: JsValue): JsResult[T] = JsSuccess(fallback)
}
def dontSerializeOption[T]: Format[Option[T]] = dontSerialize(None)

// Does not compile
case class Fur()
case class Foo(v: Option[Fur])
implicit val dontSerializeFur: Format[Option[Fur]] = dontSerializeOption[Fur]
implicit val formatFoo = Json.format[Foo]


// Workarround
case class Fur2()
type OFur2 = Option[Fur2]
case class Foo2(v: OFur2)
implicit val dontSerializeFur: Format[OFur2] = dontSerializeOption[Fur2]
implicit val formatFoo = Json.format[Foo2]
```

cchantep:


cchantep:
It generally applies the `dealias` to resolve type, and thus also fixes the handling of `Option` alias.

PR #96, opened 2017-08-25
Preserve ordering of collections in traversableWrites

### LINKED ISSUE #95: toJson does not always preserve the order of elements in collection
### Play Version 2.6.3

I found that `LowPriorityWrites.traversableWrites` does not always preserve the order.
For instance we have `scala.collection.mutable.SortedSet` of elements, however json has wrong output because of `as.map(w.writes(_))`
I suppose it should be something like `as.toSeq.map(w.writes(_)`

### Expected Behavior

```
Json.toJson(scala.collection.mutable.SortedSet(1, 2, 3)) = [1, 2, 3]
```

### Actual Behavior

```
Json.toJson(scala.collection.mutable.SortedSet(1, 2, 3)) = [2,1,3]
```

gmethvin:
Fixes #95

### DIFF
--- a/play-json/shared/src/main/scala/Writes.scala
+++ b/play-json/shared/src/main/scala/Writes.scala
@@ -301,10 +301,16 @@ sealed trait LowPriorityWrites extends EnvWrites {
   /**
    * Serializer for Traversables types.
    */
-  implicit def traversableWrites[A: Writes] = {
+  implicit def traversableWrites[A: Writes]: Writes[Traversable[A]] = {
     val w = implicitly[Writes[A]]
 
-    Writes[Traversable[A]] { as => JsArray(as.map(w.writes(_)).toArray[JsValue]) }
+    Writes[Traversable[A]] { as =>
+      val builder = mutable.ArrayBuilder.make[JsValue]()
+      as.foreach { a =>
+        builder += w.writes(a)
+      }
+      JsArray(builder.result())
+    }
     // Avoid resolution ambiguity with more specific Traversable Writes,
     // such as OWrites.map
   }
--- a/play-json/shared/src/test/scala/WritesSharedSpec.scala
+++ b/play-json/shared/src/test/scala/WritesSharedSpec.scala
@@ -6,7 +6,7 @@ package play.api.libs.json
 import org.scalatest._
 
 class WritesSharedSpec extends WordSpec with MustMatchers {
-  "Functionnal Reads" should {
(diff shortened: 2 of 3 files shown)


play-json/shared/src/test/scala/WritesSharedSpec.scala
```diff
@@ -23,6 +23,16 @@ class WritesSharedSpec extends WordSpec with MustMatchers {
     }
   }
 
+  "Traversable Writes" should {
+    "write Seqs" in {
+      Json.toJson(Seq(5, 4, 3, 2, 1)) mustEqual Json.arr(5, 4, 3, 2, 1)
+    }
+    "write SortedSets" in {
+      import scala.collection.immutable.SortedSet
+      Json.toJson(SortedSet(1, 2, 3, 4, 5)) mustEqual Json.arr(1, 2, 3, 4, 5)
```
cchantep: `immutable.SortedSet` is already ok with master for me, would rather test with a mutable one

play-json/shared/src/main/scala/Writes.scala
```diff
@@ -301,10 +301,16 @@ sealed trait LowPriorityWrites extends EnvWrites {
...
    * Serializer for Traversables types.
    */
-  implicit def traversableWrites[A: Writes] = {
+  implicit def traversableWrites[A: Writes]: Writes[Traversable[A]] = {
     val w = implicitly[Writes[A]]
 
-    Writes[Traversable[A]] { as => JsArray(as.map(w.writes(_)).toArray[JsValue]) }
+    Writes[Traversable[A]] { as =>
+      val builder = mutable.ArrayBuilder.make[JsValue]()
+      for (a <- as) {
```
cchantep: `as.foreach`? As each iteration is side-effect

play-json/shared/src/test/scala/WritesSharedSpec.scala
gmethvin: This test fails on master. I just added another one with the mutable version for completeness though.

play-json/shared/src/main/scala/Writes.scala
gmethvin: This doesn't have a `yield` so it is just side effecting, and desugars to `foreach`. That said, there is is no real advantage in readability here for `for` vs `foreach` so I'll switch to foreach.

gmethvin:
@cchantep Updated

PR #98, opened 2017-09-02
Update LICENSE and headers

gmethvin:
We should be including the whole Apache license in the LICENSE file. This way GitHub can also autodetect the license. I also set up sbt-header to update and validate the file headers.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -63,6 +63,12 @@ scalaJSStage in ThisBuild := (sys.props.get("scalaJSStage") match {
 })
 
 lazy val commonSettings = SbtScalariform.scalariformSettings ++ Seq(
+    headerLicense := {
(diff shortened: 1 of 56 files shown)


build.sbt
```diff
@@ -219,4 +225,4 @@ checkCodeFormat := {
   }
 }
 
-addCommandAlias("validateCode", ";scalariformFormat;checkCodeFormat")
+addCommandAlias("validateCode", ";scalariformFormat;headerCreate;checkCodeFormat")
```
marcospereira: Need to add `test:scalariformFormat` and `test:headerCreate` (`headerCheck;test:headerCheck`).

gmethvin: thanks

PR #102, opened 2017-09-07
Check in macro that some subclass are found for a sealed trait

cchantep:
Prevent generating useless handler for an empty sealed family.

```scala
scala> import play.api.libs.json._
import play.api.libs.json._

scala> sealed trait Foo {}
defined trait Foo

scala> Json.reads[Foo]
<console>:16: error: Sealed trait Foo is not supported: no known subclasses
       Json.reads[Foo]
```

### DIFF
--- a/play-json/shared/src/test/scala/MacroSpec.scala
+++ b/play-json/shared/src/test/scala/MacroSpec.scala
@@ -110,6 +110,10 @@ class MacroSpec extends WordSpec with MustMatchers
       val simple = Simple("foo")
       val optional = Optional(None)
 
+      shapeless.test.illTyped("Json.reads[EmptyFamily]")
+      shapeless.test.illTyped("Json.writes[EmptyFamily]")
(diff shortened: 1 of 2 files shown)


play-json/shared/src/test/scala/MacroSpec.scala
```diff
@@ -110,6 +110,10 @@ class MacroSpec extends WordSpec with MustMatchers
       val simple = Simple("foo")
       val optional = Optional(None)
 
+      shapeless.test.illTyped("Json.reads[EmptyFamily]")
```
cchantep: Before the fix:

```
[error] /Users/cchantep/Projects/play-json/play-json/shared/src/test/scala/MacroSpec.scala:113: Type-checking succeeded unexpectedly.
[error] Expected some error.
[error]       shapeless.test.illTyped("Json.reads[EmptyFamily]")
[error]
```

PR #103, opened 2017-09-08
Jsobject lookup

gpoulin:
## Purpose

Limit the amount of memory allocation during deserialization of JSON object to improve performance. This change doesn't affect the idiomaticity of the code.

## Background Context

While doing some benchmarking on our code, I found out that we were spending a non-negligible amount of time building `Map` when deserialization of JSON. By investigating deeper, I found out that play-json was doing multiple copy of the same `Map`.

For example, the following code with the current implementation, will create a `mutable.ListBuffer` convert it to `mutable.LinkedHashMap` and convert this `mutable.LinkedHashMap` to a `immutable.HashMap` before doing a lookup of the key "d".

```scala
(Json.parse("""{"a": 1, "b": 2, "c": 3, "d": 4, "e": 5}""") \ "d").as[Int]
```

We the proposed change, this code would create directly a `mutable.LinkedHashMap` and to the lookup directly on this `mutable.LinkedHashMap`.

Alternatively, we could build directly a `immutable.ListMap` which would still preserve the ordering of the keys and is likely to be faster for small JsObject (which is probably the majority of the JsObject created), but would have big penalty for big one.

### DIFF
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -12,14 +12,12 @@ import com.fasterxml.jackson.databind.Module.SetupContext
 import com.fasterxml.jackson.databind.`type`.TypeFactory
 import com.fasterxml.jackson.databind.deser.Deserializers
 import com.fasterxml.jackson.databind._
-import com.fasterxml.jackson.databind.node.{ ArrayNode, ObjectNode }
 import com.fasterxml.jackson.databind.module.SimpleModule
 import com.fasterxml.jackson.databind.ser.Serializers
 import play.api.libs.json._
 
 import scala.annotation.{ switch, tailrec }
 import scala.collection.mutable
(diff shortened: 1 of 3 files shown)


play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -108,12 +106,13 @@ private[jackson] case class ReadingList(content: mutable.ArrayBuffer[JsValue]) e
 }
 
 // Context for reading an Object
-private[jackson] case class KeyRead(content: ListBuffer[(String, JsValue)], fieldName: String) extends DeserializerContext {
```
gmethvin: You can add MiMa exclusions for these since these are `private[jackson]` and not part of the public api.

richdougherty:
@gpoulin this seems like a great idea. I think to make performance related changes we would want JMH benchmarks (you can see an example of how to add them here: https://github.com/playframework/playframework/pull/6297). Alternatively, at least some examples of performance tests that you run manually.

PR #110, opened 2017-09-26
Improve json lookup

amannocci:
## Motivation
Avoid to validate with a Reads when we already know that we fail and when we request an Option.
We don't need to compute any validation errors.

## Modification  
Just override asOpt method in JsUndefined to return a None instead of trying a validation.

## Result 
As a result we avoid useless compute and get better performance.


amannocci:
Sounds good to me.

cchantep:
@amannocci you could add a commit with suggested changes, then the review can go one. Best regards.

play-json/shared/src/main/scala/play/api/libs/json/JsLookup.scala
```diff
@@ -173,13 +156,37 @@ object JsLookupResult {
 /**
  * Wrapper for JsValue to represent an existing Json value.
  */
-case class JsDefined(value: JsValue) extends AnyVal with JsLookupResult
+case class JsDefined(value: JsValue) extends AnyVal with JsLookupResult {
+
+  override def toOption: Option[JsValue] = Some(value)
```
cchantep: `override` modifier is not required (and doesn't make sense) as there is no implementation in the parent type overridden there

play-json/shared/src/main/scala/play/api/libs/json/JsLookup.scala
```diff
@@ -173,13 +156,37 @@ object JsLookupResult {
...
+  override def validateOpt[A](implicit rds: Reads[A]): JsResult[Option[A]] = Reads.optionWithNull(rds).reads(value)
+
+}
 
 /**
  * Represent a missing Json value.
  */
 final class JsUndefined(err: => String) extends JsLookupResult {
+
+  override def toOption: Option[JsValue] = None
```
cchantep: `val`

play-json/shared/src/main/scala/play/api/libs/json/JsLookup.scala
```diff
@@ -173,13 +156,37 @@ object JsLookupResult {
...
+
+  override def toEither: Either[JsonValidationError, JsValue] = Left(validationError)
+
   override def asOpt[T](implicit fjs: Reads[T]): Option[T] = None
+
+  override def isEmpty: Boolean = true
```
cchantep: `val`

play-json/shared/src/main/scala/play/api/libs/json/JsLookup.scala
```diff
@@ -173,13 +156,37 @@ object JsLookupResult {
...
  * Wrapper for JsValue to represent an existing Json value.
  */
-case class JsDefined(value: JsValue) extends AnyVal with JsLookupResult
+case class JsDefined(value: JsValue) extends AnyVal with JsLookupResult {
+
+  override def toOption: Option[JsValue] = Some(value)
+
+  override def toEither: Either[JsonValidationError, JsValue] = Right(value)
+
+  override def isEmpty: Boolean = false
```
cchantep: `val`

play-json/shared/src/main/scala/play/api/libs/json/JsLookup.scala
```diff
@@ -125,23 +125,14 @@ sealed trait JsLookupResult extends Any with JsReadable {
   /**
```
cchantep: Please also moves the `JsLookupResult` sealed family in a separate `JsLookupResult.scala`

play-json/shared/src/main/scala/play/api/libs/json/JsLookup.scala
```diff
@@ -125,23 +125,14 @@ sealed trait JsLookupResult extends Any with JsReadable {
...
 
   def get: JsValue = toOption.get
 
   def getOrElse(v: => JsValue): JsValue = toOption.getOrElse(v)
-  def isEmpty: Boolean = this match {
-    case JsUndefined() => true
-    case JsDefined(_) => false
-  }
+  def isEmpty: Boolean
   def isDefined: Boolean = !isEmpty
```
cchantep: Would make MiMa unhappy (so not to be done right now) but making this one `final` would make sense for me, to ensure `isDefined` has the same behaviour for every impl

cchantep:
@gmethvin I would make sense to me to filter such [MiMa warnings](https://travis-ci.org/playframework/play-json/jobs/281671837#L859)

play-json/shared/src/main/scala/play/api/libs/json/JsLookup.scala
amannocci: Done in 4bdf67c

play-json/shared/src/main/scala/play/api/libs/json/JsLookup.scala
amannocci: I'm not sure but this can't be done in a value class right ?

gmethvin: there is no benefit to using a `val` here because the return value is already the constant `false`.

gmethvin:
We can filter those MiMa warnings since this is a sealed trait and we provide all the implementations. From the point of view of the user this is an internal refactoring.

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
```diff
@@ -0,0 +1,81 @@
...
+
+sealed trait JsLookupResult extends Any with JsReadable {
+  /**
+   * Tries to convert the node into a JsValue
+   */
+  def toOption: Option[JsValue]
+
+  def toEither: Either[JsonValidationError, JsValue]
+
+  def get: JsValue = toOption.get
```
cchantep: Can also be refactored in the same way (with pure `val` for `JsDefined`)

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  def getOrElse(v: => JsValue): JsValue = toOption.getOrElse(v)
+  def isEmpty: Boolean
+  def isDefined: Boolean = !isEmpty
```
cchantep: `isDefined` and `getOrElse` could be `final` to enforce implementing class respect this contract

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * If this result is defined return `this`. Otherwise return `alternative`.
+   */
+  def orElse(alternative: => JsLookupResult): JsLookupResult =
+    if (isDefined) this else alternative
```
cchantep: As conditional, could also be refactored with impl in subclasses only

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Wrapper for JsValue to represent an existing Json value.
+ */
+case class JsDefined(value: JsValue) extends AnyVal with JsLookupResult {
+
+  def toOption: Option[JsValue] = Some(value)
+
+  def toEither: Either[JsonValidationError, JsValue] = Right(value)
+
+  def isEmpty: Boolean = false
```
cchantep: Would prefer `isEmpty` `toEither` `toOption` as `val`

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
gmethvin: In `JsUndefined` this could be something like:
```scala
throw new NoSuchElementException(error)
```

which would be more useful than the generic exception thrown by `Option#get`

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  override def asOpt[T](implicit fjs: Reads[T]): Option[T] = None
+
+  val isEmpty: Boolean = true
+
+  def validate[A](implicit rds: Reads[A]): JsResult[A] = JsError(validationError)
+
+  def validateOpt[A](implicit rds: Reads[A]): JsResult[Option[A]] = JsSuccess(None)
+
+  def error = err
```
cchantep: Maybe this `def` can be removed, by refactoring the constructor as `.. class JsUndefined(val error: => String) ..`

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright (C) 2009-2017 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.api.libs.json
+
+sealed trait JsLookupResult extends Any with JsReadable {
```
gmethvin: Can we add documentation to all the trait methods here?

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
gmethvin: The same could be done for `isEmpty`, and anything else that could be implemented based on other methods here. We should pick one style (either implementing in the trait or in subclasses) and stick to it. I'm not that concerned about forcing implementing classes to do things since we control the implementations.

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
gmethvin: you can't have by-name arguments as vals.

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
gmethvin: Doesn't that just mean we're storing the value in two places now? I don't see the benefit.

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
cchantep: `isEmpty` and `isDefined` must follow the same logic, but be linked by negation, so having one depending on the type and the other always forced (`final` impl) makes sense to me, as in https://github.com/scala/scala/blob/2.12.x/src/library/scala/Option.scala#L107

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
cchantep: Had doubt about that. Any `err` and `error` look like duplicate for me.

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
cchantep: For me it just means that the different representations of the pure value are only instantiated once (e.g. no different `Some` for each call to `toOption`). Could also be `lazy val`, but I see few benefit there.

play-json/shared/src/main/scala/play/api/libs/json/JsLookupResult.scala
amannocci: We can't access `err` in JsLookup without `error`

gmethvin: Maybe you should make the constructor parameter `_error` for clarity.

amannocci:
@gmethvin The purpose of this PR was just to improve `asOpt` in JsUndefined. 
In the old style this method can be very slow compare to the new one (400% in a JMH).
And @cchantep propose to perform a global refactor on `JsLookupResult` because we can do the same on `JsLookupResult`.
Let me know what you want to finally do.
On the base I just point out that we can optimize the above method.

gmethvin:
@amannocci It would be great to add a jmh test in `benchmarks` so we can see the difference.

Another way (in the style of scala Option/Either) would be to override `asOpt` in `JsLookupResult`:
```scala
def asOpt[T](implicit fjs: Reads[T]): Option[T] = if (isEmpty) None else super.asOpt[T]
```
how much of a difference do you see with that versus your solution?

If we're going for performance here it also might be useful to swap the cases in the match statements, so we check the fast path first:

so
```scala
def toOption: Option[JsValue] = this match {
  case undef: JsUndefined => None
  case JsDefined(v) => Some(v)
}
```
and
```scala
def validate[A](implicit rds: Reads[A]): JsResult[A] = this match {
  case undef: JsUndefined => JsError(undef.validationError)
  case JsDefined(v) => v.validate[A]
}
```

cchantep:
For functions which are strictly and most dependent on the "real" type, for me in makes sense to be in these sub-types, where it can in most cases be implemented as pure val without conditional (as [Try](https://github.com/scala/scala/blob/2.12.x/src/library/scala/util/Try.scala#L62), or mostly `Option`).

Anyway would also be interested to see microbench.

amannocci:
I add a little JMH `JsonLookups_01_LookupUndef.retrieve` to check if there is a real performance improvement in all cases.
Each result is available under different branch. 

##### Branch: bench/original
```
[info] Benchmark                             Mode  Cnt       Score      Error  Units
[info] JsonLookups_01_LookupUndef.retrieve  thrpt    5  728626,530 ± 7439,065  ops/s
```
##### Branch: master
```
[info] Benchmark                             Mode  Cnt       Score      Error  Units
[info] JsonLookups_01_LookupUndef.retrieve  thrpt    5  848444,494 ± 5502,128  ops/s
```
##### Branch: bench/improve
```
[info] Benchmark                             Mode  Cnt         Score         Error  Units
[info] JsonLookups_01_LookupUndef.retrieve  thrpt    5  96473741,442 ± 1203714,171  ops/s
```
We can see an improvement between the original code and this PR but I can't explain the improvement in `bench/improve`.
I think I made a mistake somewhere in the JMH.

amannocci:
I found where I made a mistake.
We want to avoid compute of `validationError` when we use `asOpt` in `JsUndefined` because this compute imply that we create a string of the actual json object.
To avoid this we can simply use `lazy val` and I think it will be good.
New result when we avoid the computation.

##### Branch: master
```bash
[info] Benchmark                             Mode  Cnt         Score        Error  Units
[info] JsonLookups_01_LookupUndef.retrieve  thrpt    5  99852707,796 ± 720557,084  ops/s
```
Maybe we can add some benchmark to cover other lookup use case ?
@cchantep @gmethvin

cchantep:
Nice work. The more bench there is the better it is to validate such changes I guess.

amannocci:
I have added micro benchmark for all lookup cases on different branchs.
Here some result :

##### Branch: bench/original
```
[info] Benchmark                                Mode  Cnt         Score        Error  Units
[info] JsonLookups_01_LookupUndef.as           thrpt  200    223162,818 ±    581,943  ops/s
[info] JsonLookups_01_LookupUndef.asOpt        thrpt  200    736978,874 ±   2962,014  ops/s
[info] JsonLookups_01_LookupUndef.validate     thrpt  200    728214,458 ±   1939,530  ops/s
[info] JsonLookups_01_LookupUndef.validateOpt  thrpt  200  63386684,261 ± 165863,166  ops/s
[info] JsonLookups_02_LookupDef.as             thrpt  200  54317603,837 ± 182630,093  ops/s
[info] JsonLookups_02_LookupDef.asOpt          thrpt  200  51291537,242 ± 262152,804  ops/s
[info] JsonLookups_02_LookupDef.validate       thrpt  200  47501720,397 ± 523872,230  ops/s
[info] JsonLookups_02_LookupDef.validateOpt    thrpt  200  31512163,338 ± 300618,881  ops/s
```
##### Branch: master
```
[info] Benchmark                                Mode  Cnt         Score        Error  Units
[info] JsonLookups_01_LookupUndef.as           thrpt  200    223198,111 ±    860,050  ops/s
[info] JsonLookups_01_LookupUndef.asOpt        thrpt  200  97040339,444 ± 796313,854  ops/s
[info] JsonLookups_01_LookupUndef.validate     thrpt  200    771964,360 ±   2804,213  ops/s
[info] JsonLookups_01_LookupUndef.validateOpt  thrpt  200  63382788,046 ± 260387,573  ops/s
[info] JsonLookups_02_LookupDef.as             thrpt  200  47802853,941 ± 548047,862  ops/s
[info] JsonLookups_02_LookupDef.asOpt          thrpt  200  49296542,123 ± 387993,062  ops/s
[info] JsonLookups_02_LookupDef.validate       thrpt  200  58189540,846 ± 244015,862  ops/s
[info] JsonLookups_02_LookupDef.validateOpt    thrpt  200  44734348,309 ±  98563,013  ops/s
```
##### Branch: bench/improve
```
[info] Benchmark                                Mode  Cnt         Score        Error  Units
[info] JsonLookups_01_LookupUndef.as           thrpt  200    226044,402 ±   1216,069  ops/s
[info] JsonLookups_01_LookupUndef.asOpt        thrpt  200  95303524,821 ± 362315,017  ops/s
[info] JsonLookups_01_LookupUndef.validate     thrpt  200    721749,981 ±   1703,048  ops/s
[info] JsonLookups_01_LookupUndef.validateOpt  thrpt  200  63536196,062 ± 105115,529  ops/s
[info] JsonLookups_02_LookupDef.as             thrpt  200  54089105,649 ± 197206,464  ops/s
[info] JsonLookups_02_LookupDef.asOpt          thrpt  200  42731390,026 ±  53030,901  ops/s
[info] JsonLookups_02_LookupDef.validate       thrpt  200  48557903,424 ± 172930,426  ops/s
[info] JsonLookups_02_LookupDef.validateOpt    thrpt  200  31921834,288 ±  68646,003  ops/s
```
We can see an improvement in almost every case except in `LookupDef.as` where there is a regression.
I don't know why there is a regression.
If someone have an idea ?

@cchantep @gmethvin

cchantep:
No idea

PR #121, opened 2017-11-02
Strict implementation for Reads when default value is provided

lavrov:
- uses default value only if json node is empty or missing
- default value will not be used in case of failure

### DIFF
--- a/play-json/shared/src/test/scala/play/api/libs/json/JsonExtensionSpec.scala
+++ b/play-json/shared/src/test/scala/play/api/libs/json/JsonExtensionSpec.scala
@@ -705,9 +705,11 @@ class JsonExtensionSpec extends WordSpec with MustMatchers {
 
       def validateReads(fooReads: Reads[WithDefault2]) = {
         fooReads.reads(Json.obj()) mustEqual JsSuccess(WithDefault2())
+        fooReads.reads(Json.obj("a" -> JsNull)) mustEqual JsSuccess(WithDefault2())
         fooReads.reads(Json.obj("bar" -> JsNull)) mustEqual JsSuccess(WithDefault2(bar = None))
         fooReads.reads(Json.obj("a" -> "z")) mustEqual JsSuccess(WithDefault2(a = "z"))
         fooReads.reads(Json.obj("a" -> "z", "bar" -> Json.obj("b" -> "z"))) mustEqual JsSuccess(WithDefault2(a = "z", bar = Some(WithDefault1(b = Some("z")))))
+        fooReads.reads(Json.obj("a" -> 1)) mustEqual JsError(List((JsPath \ "a") -> List(JsonValidationError("error.expected.jsstring"))))
       }
 
       "by functional reads" in validateReads(functionalReads)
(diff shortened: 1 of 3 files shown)


play-json/shared/src/test/scala/play/api/libs/json/JsonExtensionSpec.scala
```diff
@@ -705,9 +705,11 @@ class JsonExtensionSpec extends WordSpec with MustMatchers {
 
       def validateReads(fooReads: Reads[WithDefault2]) = {
         fooReads.reads(Json.obj()) mustEqual JsSuccess(WithDefault2())
+        fooReads.reads(Json.obj("a" -> JsNull)) mustEqual JsSuccess(WithDefault2())
         fooReads.reads(Json.obj("bar" -> JsNull)) mustEqual JsSuccess(WithDefault2(bar = None))
         fooReads.reads(Json.obj("a" -> "z")) mustEqual JsSuccess(WithDefault2(a = "z"))
         fooReads.reads(Json.obj("a" -> "z", "bar" -> Json.obj("b" -> "z"))) mustEqual JsSuccess(WithDefault2(a = "z", bar = Some(WithDefault1(b = Some("z")))))
+        fooReads.reads(Json.obj("a" -> 1)) mustEqual JsError(List((JsPath \ "a") -> List(JsonValidationError("error.expected.jsstring"))))
```
cchantep: Even if I tend to agree with that, that's a behaviour change that could "surprise" the project using the current version.

lavrov: On the other hand the current version is a "surprise". It swallows errors by providing default values which is quite annoying if you have built rest api with play-json :)

cchantep: It "can" be a surprise for new users, but many more current user can have their code relying on the current behaviour.

gmethvin: This is kind of unfortunate, because one would typically expect the default value would be applied only for null values. Maybe we need to introduce new options "default on error" and "default on null/empty".

cchantep: Yes would be better

richdougherty: 👍  to new methods with clearer semantics. I agree the current method's semantics are confusing. Maybe deprecate it once there are alternatives?

gmethvin: Okay, I actually think we should fix this "bug" for a few reasons:

 1. The current behavior is new and undocumented, so it probably is not in wide use. Even without clear documentation we've received several reports of incorrect behavior. The whole point is to be able to omit fields from the JSON object for brevity, so it would be weird for a user to expect the default-on-error behavior. 
 2. `Reads.nullableWithDefault` actually has a behavior similar to what we expect. Passing an actual value that is invalid results in `JsError`. This method is documented but the documentation needs to be updated to explain the default is used if the node is not found, rather than returning `JsError`. Given the inconsistency between these two methods, it would make sense to fix one of them.
 3. If a user assumes that the default is only applied on null or empty, the current behavior can lead to a potentially serious bug in their application. On the other hand, if we update this code and a user is assuming the "bug" behavior, then validation will fail and the problem can be easily detected. I think it is much more likely we're being helpful to existing users by making this change.
 4. To properly add additional APIs would be a breaking change anyway due to binary compatibility, unless we add the APIs in an unusual place (i.e. not in traits). We would be going through a lot of trouble and complicating the API to provide a "feature" that no one asked for.

So I think we should merge this. The only other thing I would suggest @lavrov is that we update the documentation for `withDefault` and `nullableWithDefault` and the equivalent methods on `JsPath` to properly explain the behavior.

PR #126, opened 2017-11-27
update sbt plugins

xuwei-k:


build.sbt
```diff
@@ -66,7 +66,14 @@ scalaJSStage in ThisBuild := (sys.props.get("scalaJSStage") match {
   case _ => FastOptStage
 })
 
-lazy val commonSettings = scalariformSettings(autoformat = true) ++ Seq(
+lazy val commonSettings = SbtScalariform.projectSettings ++ Seq(
+    publishTo := Some(
```
gmethvin: Are you sure we need these settings? This should be handled by the `PlayRootProject`/`PlayLibrary` plugins: https://github.com/playframework/interplay/blob/master/src/main/scala/interplay/PlayBuildBase.scala#L63

xuwei-k: https://github.com/xerial/sbt-sonatype/tree/2.0#buildsbt

`PlayLibrary` plugin does not set `publishTo` setting. We need explicit `publishTo` setting since sbt-sonatype 2.0.
`publishTo` is `None` if without this setting.

gmethvin: ok then we probably should update sbt-sonatype in interplay and add the settings there. It's still using 1.1: https://github.com/playframework/interplay/blob/master/project/plugins.sbt#L11. I guess it works, but it's safer not to mix major versions.

marcospereira: Yep. We should probably, at some point, cut a 2.0.x release of interplay. I think that upgrading to sbt 1.0 is where we will do that. And then we could upgrade plugins too.

marcospereira: > I guess it works, but it's safer not to mix major versions.

I think it is ok to update plugins here because it has no impact on users of play-json.

PR #128, opened 2017-11-29
Adding PascalCase to JsonConfiguration default implementation.

faccuo:
## Purpose

Adding PascalCase to JsonConfiguration default implementation.

## Background Context

Found it useful when reading JSON from AWS http sns-webhook since it is using pascal case.


faccuo:
I did sign the CLA.

PR #137, opened 2018-01-20
Configure macro to allow it to output null in json during writes

vdebergue:
# Pull Request Checklist

## Purpose

Add configuration to the Json macro to support cases like [this](https://stackoverflow.com/questions/34317253/explicitly-output-json-null-in-case-of-missing-optional-value)

## Background Context

I first tried to add a simple Boolean but the behaviour was not really clear in the case of `Json.format`. This approach with `OptionHandlers` makes it clearer what is used and it can be extended by a user.

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsonConfiguration.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsonConfiguration.scala
@@ -131,7 +131,9 @@ object JsonNaming {
 trait OptionHandlers {
   def readHandler[T](jsPath: JsPath)(implicit r: Reads[T]): Reads[Option[T]]
   def writeHandler[T](jsPath: JsPath)(implicit writes: Writes[T]): OWrites[Option[T]]
-  def formatHandler[T](jsPath: JsPath)(implicit format: Format[T]): OFormat[Option[T]]
+  final def formatHandler[T](jsPath: JsPath)(implicit format: Format[T]): OFormat[Option[T]] = {
+    OFormat(readHandler(jsPath), writeHandler(jsPath))
+  }
 }
 
 /** OptionHandlers companion */
@@ -139,12 +141,11 @@ object OptionHandlers {
 
   /**
    * Default Option Handlers
-   * Uses readNullable and writesNullable under the hood
+   * Uses readNullable and writesNullable
    */
   object Default extends OptionHandlers {
     def readHandler[T](jsPath: JsPath)(implicit r: Reads[T]): Reads[Option[T]] = jsPath.readNullable
     def writeHandler[T](jsPath: JsPath)(implicit writes: Writes[T]): OWrites[Option[T]] = jsPath.writeNullable
-    def formatHandler[T](jsPath: JsPath)(implicit format: Format[T]): OFormat[Option[T]] = jsPath.formatNullable
   }
 
   /**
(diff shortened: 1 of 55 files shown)


play-json/shared/src/main/scala/play/api/libs/json/JsonConfiguration.scala
```diff
@@ -118,3 +126,36 @@ object JsonNaming {
...
+
+/** OptionHandlers companion */
+object OptionHandlers {
+
+  /**
+   * Default Option Handlers
+   * Uses readNullable and writesNullable under the hood
+   */
+  object Default extends OptionHandlers {
+    override def readHandler[T](jsPath: JsPath)(implicit r: Reads[T]): Reads[Option[T]] = jsPath.readNullable
```
cchantep: Would remove `override` modifier there

play-json/shared/src/main/scala/play/api/libs/json/JsonConfiguration.scala
```diff
@@ -11,15 +11,21 @@ sealed trait JsonConfiguration {
...
+  def optionHandlers: OptionHandlers = OptionHandlers.Default
 }
 
 object JsonConfiguration {
   type Aux[O <: Json.MacroOptions] = JsonConfiguration { type Opts = O }
 
   private final class Impl[O <: Json.MacroOptions](
-    val naming: JsonNaming = JsonNaming.Identity
+    val naming: JsonNaming = JsonNaming.Identity,
+    override val optionHandlers: OptionHandlers = OptionHandlers.Default
```
cchantep: Would not put `override` modifier there

vdebergue: The modifier is necessary here:
`overriding method optionHandlers in trait JsonConfiguration of type => play.api.libs.json.OptionHandlers`

I wanted to not set `optionHandlers` in the trait but then it failed the binary compatibility check.

cchantep: As the implementation is provided in the trait as `OptionHandlers.Default`, and as this `Impl` class is not part of the public API (as `private`), I would add a filter in MiMa (would found it cleaner than `override`).

vdebergue: Binary compatibility is hard, I'll take an other look at it later

cchantep: You can copy the warning when there is no such `override` and paste it in the MiMa filters already in `build.sbt` for `play-json` module

cchantep: Would also add `ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.libs.json.JsonConfiguration.optionHandlers")` as having a new function is not breaking the API compat for the existing projects using Play JSON.

play-json/shared/src/main/scala/play/api/libs/json/JsonConfiguration.scala
```diff
@@ -118,3 +126,36 @@ object JsonNaming {
     def apply(property: String): String = transformation(property)
   }
 }
+
+/** Configure how options should be handled */
+trait OptionHandlers {
```
cchantep:
```scala
/**
 * Returns a `OFormat` instance based on the current read and write handlers for `Option`.
 */
final def formatHandler[T: Format](jsPath: JsPath): OFormat[Option[T]] = OFormat(readHandler[T](jsPath), writeHandler[T](jsPath))
```

PR #143, opened 2018-02-14
Fix #93: support object in macro

### LINKED ISSUE #93: json.Json.format does not support case object
this is the upstream bug for https://github.com/fommil/stalactite/issues/14

I've created a macro that makes it a lot easier to derive typeclasses with semi-auto derivation. But it seems that `.format` does not support `case object`, requiring special handling. I'd like you to consider supporting `case object` in your macro to avoid the need for special casing in stalactite. Everything based on shapeless derivation is able to support `case object` as a product alongside `case class`es, so I think play-json is the odd one out.

vdebergue:
# Pull Request Checklist

## Fixes

Fixes #93 

Scala `object`s (and `case object`s)  are serialized as empty Json objects.

I didn't make the distinction between `object` and `case object` as it seemed unecessary

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
@@ -783,7 +783,7 @@ import scala.reflect.macros.blackbox
             case _ => $json.JsError("error.expected.jsobject")
           }
          """
-      def writer = q"$json.OWrites[$atpe]{_ => $json.Json.obj() }"
+      def writer = q"$json.OWrites[$atpe]{_ => $json.JsObject.empty }"
 
       val tree = methodName match {
         case "read" => reader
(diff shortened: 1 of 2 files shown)


play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -783,11 +775,39 @@ import scala.reflect.macros.blackbox
...
+
     // ---
 
     directKnownSubclasses match {
       case Some(subTypes) => macroSealedFamilyImpl(subTypes)
-      case _ => macroCaseImpl
+      case _ =>
+        atpe match {
+          case TypeRef(_, _, args) => macroCaseImpl(args)
+          case _: SingletonType => caseObjectImpl
```
cchantep: Would move this case juste before the `TypeRef` one (more specific to me)

play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -783,11 +775,39 @@ import scala.reflect.macros.blackbox
...
 
+    def caseObjectImpl: c.Expr[M[A]] = {
+      def reader =
+        q"""
+          $json.Reads[${atpe}] {
+            case obj @ $json.JsObject(_) => $json.JsSuccess(${atpe.termSymbol})
+            case _ => $json.JsError("error.expected.jsobject")
+          }
+         """
+      def writer = q"$json.OWrites[$atpe]{_ => $json.Json.obj() }"
```
cchantep: Would rather use `JsObject.empty` there

PR #145, opened 2018-02-14
Use sbt 1.1.0 to build

marcospereira:
## Purpose

Build play-json using sbt 1.1.0.


marcospereira:
Avoid merging this. Just notice that we a fix for `docs` sub-project (we are not testing it in our builds).

ignasi35:
Raised https://github.com/playframework/play-json/issues/237

PR #147, opened 2018-02-15
Customizable type discriminators for sealed case class family macro

dwickern:
Currently the macro uses a `_type` field with the class name, ie.
```json
{ "_type" : "fully.qualified.ClassName" }
```

This change allows both the name and value to be customized.
```scala
sealed trait Base
@Discriminator("foo") case class Foo(x: Int) extends Base
@Discriminator("bar") case class Bar(y: String) extends Base

implicit val cfg = JsonConfiguration(discriminator = "_discriminator")
```

In this example the JSON would look like
```json
{ "_discriminator": "foo", "x": 42 }
```

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
@@ -535,7 +535,7 @@ import scala.reflect.macros.blackbox
                   s"No instance of Reads is available for ${t.typeSymbol.fullName} in the implicit scope (Hint: if declared in the same file, make sure it's declared before)"
                 )
               }
-              cq"name if name == $config.classNaming(${t.typeSymbol.fullName}) => $reader.reads(vjs)" :: out
+              cq"name if name == $config.typeNaming(${t.typeSymbol.fullName}) => $reader.reads(vjs)" :: out
             }
           )
         ).reverse)
@@ -577,7 +577,7 @@ import scala.reflect.macros.blackbox
               case jsv => $json.JsObject(Seq("_value" -> jsv))
             }
 
-            jso + ($config.discriminator -> $json.JsString($config.classNaming(${t.typeSymbol.fullName})))
+            jso + ($config.discriminator -> $json.JsString($config.typeNaming(${t.typeSymbol.fullName})))
           }"""
         })
 
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
@@ -519,7 +519,14 @@ import scala.reflect.macros.blackbox
         c.abort(c.enclosingPosition, s"Sealed trait ${atpe} is not supported: no known subclasses")
       }
 
-      val typeNaming = (_: Type).typeSymbol.fullName
+      def typeNaming(t: Type): String = {
+        /** Extract "name" from @Discriminator("name") */
+        val annotatedName = t.typeSymbol.annotations.map(_.tree).collectFirst {
+          case Apply(Select(New(tt), TermName("<init>")), List(Literal(Constant(name: String)))) if tt.tpe == typeOf[play.api.libs.json.Discriminator] => name
+        }
+
+        annotatedName getOrElse t.typeSymbol.fullName
+      }
 
       def readLambda: Tree = {
         val resolver = new ImplicitResolver({ orig: Type => orig })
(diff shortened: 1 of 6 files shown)


gmethvin:
You can add the mima filters since it's a sealed trait. This should be backportable to 2.6.x

dwickern:
Is there some documentation on how to build the docs?
```
dwickern:docs dwickern$ sbt -Dplay.version=2.6.11 test
[info] Loading settings from idea.sbt ...
[info] Loading global plugins from /Users/dwickern/.sbt/1.0/plugins
[info] Loading settings from plugins.sbt ...
[info] Loading project definition from /Users/dwickern/code/play-json/docs/project
[info] Loading settings from build.sbt ...
[info] Set current project to docs (in build file:/Users/dwickern/code/play-json/docs/)
[info] Updating {file:/Users/dwickern/code/play-json/docs/}docs...
[error] java.lang.IllegalArgumentException: Cannot add dependency 'com.typesafe.play#play-docs_2.12;2.6.11' to configuration 'docs' of module docs#docs_2.12;0.1-SNAPSHOT because this configuration doesn't exist!
```

play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -519,7 +519,14 @@ import scala.reflect.macros.blackbox
         c.abort(c.enclosingPosition, s"Sealed trait ${atpe} is not supported: no known subclasses")
       }
 
-      val typeNaming = (_: Type).typeSymbol.fullName
+      def typeNaming(t: Type): String = {
+        /** Extract "name" from @Discriminator("name") */
```
cchantep: I would rather have a `subtypeFqdn: String => discriminatorValue: String` as a naming strategy that can be specified thanks to `JsonConfiguration`.

It allows to manage differents namings for a dame sealed family according different use cases.

Also to have all the settings at once/in one place with the `JsonConfiguration`.

> For me it can be considered later whether to automatically generate a default naming strategy for discriminator values of types according annotation.

dwickern: I did consider this. It's useful if the discriminators can be derived from the class name, e.g. `fully.qualified.CaseClassName` => `"case-class-name"`. For my use-case, I prefer the annotation over a nasty pattern match with no exhaustiveness/correctness checking.

But I do agree with you. For a consistent API we should support the same options for class names as we have for the class member names. What do you think about this proposal?

1. Support `JsonNaming` for the class name:
```scala
implicit val c = JsonConfiguration(
  naming = JsonNaming.SnakeCase,
  classNaming = JsonNaming.SnakeCase,
  discriminator = "_disc",
)
trait Base
case class MyCaseClass(myClassMember: Int) extends Base
// { "_disc": "my_case_class", "my_class_member": 42 }
```

2. Support compile-time annotations for both class names and member names, with the annotation taking priority over other naming configuration:
```scala
trait Base
@Discriminator("custom") case class MyCaseClass(@Name("mbr") myClassMember: Int) extends Base
// { "_type": "custom", "mbr": 42 }
```

cchantep: I'm more for option 1, with everything being configurable homogeneously in `JsonConfiguration` with `JsonNaming` strategies.

Annotation can be discussed worked with a distinct PR.

dwickern:
I updated the PR to use `JsonNaming` to map between class name and discriminator value. It doesn't quite work as you'd expect because:
```scala
JsonNaming.SnakeCase("fully.qualified.ClassName") // "fully.qualified._class_name"
```

The simplest solution would be pass the class name without package, so that `"ClassName" => "class_name". But it would be a breaking change.

Alternatively a `JsonClassNaming` could be added which strips the package name:
```scala
JsonClassNaming.SnakeCase("fully.qualified.ClassName") // "class_name"
```

play-json/shared/src/main/scala/play/api/libs/json/JsonConfiguration.scala
```diff
@@ -9,9 +9,12 @@ sealed trait JsonConfiguration {
   /** Compile-time options for the JSON macros */
   type Opts <: Json.MacroOptions
 
-  /** Naming strategy */
+  /** Naming strategy for class members */
```
cchantep: for fields

play-json/shared/src/main/scala/play/api/libs/json/JsonConfiguration.scala
```diff
@@ -9,9 +9,12 @@ sealed trait JsonConfiguration {
...
   def naming: JsonNaming
 
+  /** Naming strategy for class names */
+  def classNaming: JsonNaming
```
cchantep: `typeNaming`

cchantep:
I don't see the point... the naming must be applied on the full qualified type for me.

cchantep:
@dwickern https://github.com/cchantep/play-json/commit/84dd93f67ed47d5b1287fdd37855db38cd726f69 ?

dwickern:
Thanks, I cherry picked into this PR

PR #150, opened 2018-02-25
Refactory of the dependencies. Added support and tests for java.time …

aparo:
…classes in scala.js

# Pull Request Checklist

* [X] Have you signed the [Typesafe CLA](https://www.typesafe.com/contribute/cla)?
* [X] Have you updated the documentation?

## Fixes

No issues are created, but java.time was not available in scala.js. Now all the java.time classes are serializable/deserializable in both JS and JVM.

## Purpose

The current state of scala.js components are not usable because there is not support for date/time types.
This pull request allows to use java.time classes in both worlds.

## Background Context

I added a new dependency to support java.time in scala.js in the build.sbt.
```
lazy val `play-jsonJS` = `play-json`.js.
  settings(
    libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % "2.0.0-M12"
  )
```
I fixed build.sbt also adding %%% to al common settings for js/jvm to be sure to have the right dependencies.

The previous EnvReads and EnvWrites were refactored in this way:
- the common (JS/JVM) serializers are moved in shared folder in EnvReads/EnvWrites and they are extending PlatformReads/PlatformWrites
- PlatformReads/PlatformWrites contain the serializers/deserializers that are custom for JS or JVM.

The code is unchanged the java.time serializers/deserializers are moved in EnvReads/EnvWrites because now they are shared.

The refactory of the tests is similar to EnvReads/EnvWrites. All the common tests for java.time are moved in shared so that they are common to both worlds.

## References

No references


play-json/jvm/src/main/scala/play/api/libs/json/PlatformReads.scala
```diff
@@ -0,0 +1,245 @@
...
+ */
+
+package play.api.libs.json
+import java.util.Locale
+
+import com.fasterxml.jackson.databind.JsonNode
+import com.fasterxml.jackson.databind.node.{ ArrayNode, ObjectNode }
+import play.api.libs.json.jackson.JacksonJson
+
+trait PlatformReads {
```
cchantep: I don't see the point of adding an extra level of inheritance. In master, the shared DefaultReads extends EnvReads for env specific bindings. For me these 2 levels are sufficient.

play-json/shared/src/test/scala/play/api/libs/json/WritesSpec.scala
```diff
@@ -4,20 +4,8 @@
...
-  LocalDate,
-  LocalTime,
-  OffsetDateTime,
-  ZonedDateTime,
-  ZoneOffset,
-  ZoneId
-}
 import java.time.temporal.ChronoUnit
-import java.time.format.DateTimeFormatter
+import java.time.{ Instant, LocalDate, LocalDateTime, LocalTime, OffsetDateTime, Period, ZoneId, ZoneOffset, ZonedDateTime, Duration => JDuration }
```
cchantep: Please keep the import formatting (too long line there for me)

play-json/jvm/src/main/scala/play/api/libs/json/PlatformReads.scala
aparo: The extra level is required to allows to keep the JVM only serializers/Deserializers separated by the shared ones. I  refactored in common place a lot on serializers/deserializers used for java.time to be used both in JVM and JS

cchantep: For me the separation between shared/common and specific js/JVM is already there on master so still don't see the point

PR #153, opened 2018-04-01
Added test for dynamically generated case class default values

alexkuz:
# Pull Request Checklist

## Purpose

Add test for dynamically generated case class default values
Usecase: automatically generated model id if no id is provided in JSON - should be unique for every JSON reading.

## Background Context

Since it didn't work in v0.6.7 but works now I thought it worth adding a test for this case.


cchantep:
Closing it for now.

alexkuz:
Sorry, missed your response - the point was to make sure that function is called on each deserialization. In previous versions function was called just once and it's return value was used for every generated instance, so they all had the same value (just to make things clear; if you think that's redundant - it's up to you) :)

PR #157, opened 2018-05-04
Support ValueClass in macros

cchantep:


play-json/shared/src/main/scala/play/api/libs/json/Json.scala
```diff
@@ -248,6 +248,24 @@ object Json extends JsonFacade {
...
+   *
+   * {{{
+   * import play.api.libs.json.{ Json, Reads }
+   *
+   * final class IdText(val value: String) extends AnyVal
+   *
+   * val r: Reads[IdText] = Json.valueReads
+   * }}}
+   */
+  def valueReads[A]: Reads[A] = macro JsMacroImpl.implicitConfigValueReads[A]
```
cchantep: @richdougherty If we intend that to be only in 2.7.0, I think it may be possible to update the `def reads` to have 2 unambiguous variants (able to infer whether to call the ValueClass macro or not), only using `reads[T]` at the calling site.

play-json/shared/src/main/scala/play/api/libs/json/Json.scala
```diff
@@ -248,6 +248,24 @@ object Json extends JsonFacade {
    */
   def reads[A]: Reads[A] = macro JsMacroImpl.implicitConfigReadsImpl[A]
 
+  /**
+   * Creates a `Reads[A]`, if `A` is a ValueClass,
+   * by resolving at compile-time the required implicits.
```
gmethvin: Can we explain what it does here? e.g. "by using the `Reads` of the underlying type". This is the behavior I expected but it might not be 100% obvious to everyone.

play-json/shared/src/main/scala/play/api/libs/json/Json.scala
gmethvin: It would be nice to get this into 2.6.x, so maybe we keep it like this and we can improve upon `reads` in 2.7?

cchantep:
(Documented)[https://github.com/playframework/play-json/pull/157/files#diff-e51f048c1e281a2f6277eb9bc82d985fR12]

docs/manual/working/scalaGuide/main/json/ScalaJsonAutomated.md
```diff
@@ -9,6 +9,14 @@ The following macro will create a `Reads[Resident]` based on its structure and t
...
+The [value classes](https://docs.scala-lang.org/overviews/core/value-classes.html) are also supported. Given the following value class, based on a `String` value:
+
+@[valueClass](code/ScalaJsonAutomatedSpec.scala)
+
+Then it's also possible to generate a `Reads[IdText]` using the following macro (as `String` is already supported):
+
+@[value-reads](code/ScalaJsonAutomatedSpec.scala)
+
 When compiling, the macro will inspect the given class and
 inject the following code, exactly as if you had written it manually:
```
gmethvin: The way this is written now it looks like this is could be about value classes. I'd move the value class stuff to a different section, and explicitly say that a value class will be serialized exactly as the wrapped type is serialized.

cchantep:
Updated

PR #78, opened 2017-07-03
Scala 2.13.0-M1

xuwei-k:
Fixes https://github.com/playframework/play-json/issues/61

PR #158, opened 2018-05-10
Fix #79 - Applicative .pure laziness

cchantep:


play-json/shared/src/main/scala/play/api/libs/json/Reads.scala
```diff
@@ -98,14 +98,26 @@ object Reads extends ConstraintReads with PathReads with DefaultReads with Gener
...
+   * import play.api.libs.json.Reads
+   *
+   * val r: Reads[String] = Reads.pure(f = "foo")
+   * // NOTE the required `f =` to pass as named parameter
+   * }}}
+   */
+  override def pure[A](f: => A): Reads[A] = Reads[A] { _ => JsSuccess(f) }
+
+  @deprecated("Use `pure` with `f:=>A` parameter", "2.7.0")
+  def pure[A](value: A): Reads[A] = Reads[A] { _ => JsSuccess(value) }
```
richdougherty: So keeping the old version of `pure` around will help with binary compatibility, but it requires people to use `pure(f = ...)` to get the new version of `pure`?

Personally, I think it should be OK to remove the old version of `pure` and just use the new one, even if it's not binary compatible, since this change will happen for the next 'major' release of play-json anyway. For major releases, I think that source-compatibility is enough.

play-json/shared/src/test/scala/play/api/libs/json/JsonExtensionSpec.scala
```diff
@@ -744,9 +744,9 @@ class JsonExtensionSpec extends WordSpec with MustMatchers {
       }
 
       "by functional reads" in validateReads(functionalReads)
-      "by functional formats" in validateReads(functionalFormat)
-      "by reads macro" in validateReads(macroReads)
-      "by format macro" in validateReads(macroFormat)
+      //"by functional formats" in validateReads(functionalFormat)
+      //"by reads macro" in validateReads(macroReads)
+      //"by format macro" in validateReads(macroFormat)
```
richdougherty: Remove comments?

play-json/shared/src/test/scala/play/api/libs/json/MacroSpec.scala
```diff
@@ -327,6 +327,12 @@ class MacroSpec extends WordSpec with MustMatchers
       )
     }
 
+    "handle case class with default values, format defined in companion object" in {
+      val json = Json.obj("id" -> 15)
+      val expected = WithDefaultInCompanion(15, "a")
+      Json.fromJson[WithDefaultInCompanion](json).get mustEqual expected
+    }
```
richdougherty: Is this related to the `pure` change? If it's just a new test, that's OK, but I want to make sure I understand it if it's somehow testing something to do with `pure`.

cchantep: That's the tests failing at #78 due to the invalid laziness

play-json/shared/src/main/scala/play/api/libs/json/Reads.scala
gmethvin: I think we can do a 2.7.x release of play-json at some point. Not sure when though. WDYT @richdougherty?

What about using `private[json]` on this one to maintain bincompat?

cchantep: done

play-json/shared/src/test/scala/play/api/libs/json/MacroSpec.scala
cchantep: @richdougherty updated

PR #160, opened 2018-05-23
[doc] document default values option

bomgar:

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,18 @@ implicit val residentFormat = Json.format[Resident]
 
 With the `Reads` and/or `Writes` in scope, I can then easily convert my class using `toJson` and `fromJson`
 
+In case you don't want to use an Option but a sensible default value you can use case class default parameters
+
+```scala
+case class Resident(name: String, age: Int, role: Role = Role.DefaultRole)
+```
+
+You need to enable DefaultValues for this to work
+```scala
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -174,6 +174,18 @@ implicit val residentFormat = Json.format[Resident]
 
 With the `Reads` and/or `Writes` in scope, I can then easily convert my class using `toJson` and `fromJson`
 
+In case you don't want to use an Option but a sensible default value you can use case class default parameters
```
cchantep:
```
an `Option`
```

cchantep: ", you can define default values for case class."

README.md
```diff
@@ -174,6 +174,18 @@ implicit val residentFormat = Json.format[Resident]
...
+
+```scala
+case class Resident(name: String, age: Int, role: Role = Role.DefaultRole)
+```
+
+You need to enable DefaultValues for this to work
+```scala
```
cchantep: Blank line before code

README.md
```diff
@@ -174,6 +174,18 @@ implicit val residentFormat = Json.format[Resident]
...
+implicit val residentFormat: Format[Resident] = Json.using[Json.WithDefaultValues].format[Resident]
+```
+
```
cchantep: Only one blank line

README.md
bomgar: That doesn't sound right

bomgar:
Fine i won't contribute.

PR #161, opened 2018-05-28
Make JsObject.underlying public

nadavwr:
Having a private case class field interferes with automatic typeclass derivation for both Shapeless and Magnolia.
Exposing JsObject's (possibly mutable) internal structure is consistent with JsArray.

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala
@@ -121,9 +121,7 @@ object JsArray extends (IndexedSeq[JsValue] => JsArray) {
 /**
  * Represent a Json object value.
  */
-case class JsObject(
-  private[json] val underlying: Map[String, JsValue]
-) extends JsValue {
+case class JsObject(underlying: Map[String, JsValue]) extends JsValue {
 
   /**
    * The fields of this JsObject in the order passed to to constructor


gmethvin:
I'm curious when you want to use automatic typeclass generation on a `JsObject` though. A `JsObject` is a special kind of case class so typically I'd want a custom implementation (e.g. I'd want a `Show` instance that prints in JSON format).

I don't feel that strongly either way here. If we do this, what about changing the name of `underlying` to something more meaningful, like `fieldsMap`?

cchantep:
There are already `.fields` and `.value` so what?

gmethvin:
Yeah, I'm not sure I understand the use case of automatic typeclass derivation for JSON AST types. I feel like the fact that these things are case classes is more of an implementation detail. It's not a case class in the traditional sense.

nadavwr:
A bit late to the game, here are two use cases we encountered.

The JSON AST is *almost* a straightforward ADT. Getting it there would be trivial. This is a non-feature as far as play-json is concerned, but would make life much easier on users with ADT's that happen to have `JsValue` members (e.g. `case class MessageEnvelope(payload: JsValue, ...)`). This would help typeclass derivation (and other ADT-friendly macros) just work for such cases, with no need for library bindings.

1) Internal validation library 

We use an internally developed validation library, and rely on Shapeless for typeclass derivation. Some case classes we validate have `JsValue` members, so typeclass derivation fails unless we define default instances for `JsValue`. This is obviously no big deal, but from the perspective of a user adopting JSON library X and validation library Y, having them able to just work together would be nice.

2) Quicklens

We were frustrated with the JSON transformation API. At the time we also made use of [Quicklens](https://github.com/adamw/quicklens) for data sanitation over case classes, so it was natural for us to try using it with `JsValue`.

Quicklens is nowhere near the richness and depth of Monocle, but it's small, self contained, and very easy to adopt. It is primarily geared towards standard library data structures and ADT's, and isn't very extensible. There's no actual typeclass derivation involved in Quicklens' lens generation, but the macros involved fail to work with play-json only because of `JsObject`'s private field, so I'm bundling this under the same subject.

With a public `underlying` field, Quicklens works fine. The syntax ends up being a bit awkward, but it's not far from being a good fit (a few improvements are possible on Quicklens' side, I won't go into details).

cchantep:
As said before "there are already .fields and .value so what?" ... I'm far from fond of exposing a mutable internal field.

nadavwr:
`.fields` and `.value` do not help typeclass derivation. `.underlying` is part of the product type, and keeping it private breaks automatic typeclass derivation—macros used for generating a generic representation of `JsObject` would fail.

I wouldn't be fond of exposing mutable internal state either, but for the JSON AST that's already the case (`JsArray.value` is a public, possibly mutable  `scala.collection.IndexedSeq`). Might as well make both consistent (either way).

I suppose a generic representation of the AST can be provided manually (e.g. using `.value` to work around the private member). This too would enable typeclass derivation, at least for shapeless-based libraries.

cchantep:
Sorry but you still don't explain make a point why `.fields` and `.value` are not sufficient.
A lib not to derive/use that is rather the issue for me.

PR #178, opened 2018-09-06
Add `exists` & `contains` methods for `JsResult`

aaabramov:
# Pull Request Checklist

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsResult.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsResult.scala
@@ -207,6 +207,12 @@ sealed trait JsResult[+A] { self =>
     def withFilter(q: A => Boolean) = new WithFilter(a => p(a) && q(a))
   }
 
+  /** If this result is successful than checks for presence for '''elem''', otherwise return '''false''' */
+  def contains[AA >: A](elem: AA): Boolean = asOpt.contains(elem)
+
+  /** If this result is successful than check value with predicate '''p''', otherwise return '''false''' */
+  def exists(p: A => Boolean): Boolean = asOpt.exists(p)
+
   /** Updates the JSON path */
   def repath(path: JsPath): JsResult[A]
 
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsResult.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsResult.scala
@@ -21,6 +21,10 @@ case class JsSuccess[T](value: T, path: JsPath = JsPath()) extends JsResult[T] {
 
   def foreach(f: T => Unit): Unit = f(value)
 
+  def contains[AA >: T](elem: AA): Boolean = elem == value
+
+  def exists(p: T => Boolean): Boolean = p(value)
+
   def repath(path: JsPath): JsResult[T] = JsSuccess(value, path ++ this.path)
 
   def getOrElse[U >: T](t: => U): U = value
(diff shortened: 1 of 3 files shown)


aaabramov:
Any suggestions how to properly add new methods to `play-json`?
```
[error] * method contains(java.lang.Object)Boolean in trait play.api.libs.json.JsResult is present only in current version
[error]   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.libs.json.JsResult.contains")
[error] * method exists(scala.Function1)Boolean in trait play.api.libs.json.JsResult is present only in current version
[error]   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.libs.json.JsResult.exists")
[info]play-json: found 2 potential binary incompatibilities while checking against com.typesafe.play:play-json_2.11:2.6.0  (filtered 17)
[error] * method contains(java.lang.Object)Boolean in trait play.api.libs.json.JsResult is present only in current version
[error]   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.libs.json.JsResult.contains")
[error] * method exists(scala.Function1)Boolean in trait play.api.libs.json.JsResult is present only in current version
[error]   filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.libs.json.JsResult.exists")
```

aaabramov:
@dwijnand could you help with this?

renatocaval:
@aaabramov, you need to add the suggested filters on this part of the build configuration.

https://github.com/playframework/play-json/blob/master/build.sbt#L123-L144

It's safe to add the exclusions because `JsResult` is a sealed trait and therefore we know the existing implementations.

play-json/shared/src/main/scala/play/api/libs/json/JsResult.scala
```diff
@@ -207,6 +207,12 @@ sealed trait JsResult[+A] { self =>
     def withFilter(q: A => Boolean) = new WithFilter(a => p(a) && q(a))
   }
 
+  /** If this result is successful than checks for presence for '''elem''', otherwise return '''false''' */
+  def contains[AA >: A](elem: AA): Boolean = asOpt.contains(elem)
```
cchantep: Rather move implementation in the subtypes, as for the other functions.

play-json/shared/src/test/scala/play/api/libs/json/JsResultSpec.scala
```diff
@@ -48,4 +48,37 @@ class JsResultSpec extends WordSpec with MustMatchers {
       JsError("err").recoverWith { _ => succ } mustEqual succ
     }
   }
+
+  "JsResult" should {
+
+    "return true for JsSuccess(x)#contains(x)" in {
+      val x = 10
```
cchantep: Do not need such trivial `val`

aaabramov:
@cchantep, what about adding `forall` with `Option` semantics?

play-json/shared/src/test/scala/play/api/libs/json/JsResultSpec.scala
aaabramov: @cchantep I don't agree with you. I think consistent test values and absence of magic numbers is good for tests.
If you would like to, I can rewrite tests with table driven cases.
Moreover, you can do it yourself :)

play-json/shared/src/main/scala/play/api/libs/json/JsResult.scala
aaabramov: Done.

play-json/shared/src/test/scala/play/api/libs/json/JsResultSpec.scala
cchantep: There I see no value to declared such expected result as `val` as it's pure/simple/literal.

aaabramov: Do you think it is better to have
`assert(JsSuccess(10).contains(10))`?

aaabramov:
@cchantep done.

cchantep:
Thx, either add `forall` there or in a separate PR.

aaabramov:
@cchantep `forall` added in scope of https://github.com/playframework/play-json/pull/184

PR #181, opened 2018-09-18
don't accept inexact numbers while parsing date values, closes #180

### LINKED ISSUE #180: Denial of service parsing JsNumber as Instant/ZonedDateTime/LocalDateTime

2.6.9

Scala 2.12.6

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)

MacOS 10.13.6

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

Paste the output from `java -version` at the command line.

### Library Dependencies

N/A

### Expected Behavior

Run the following code:
```scala
  val number = BigDecimal(BigInt(Long.MaxValue), Int.MaxValue, MathContext.DECIMAL128)
  val jsNumber = JsNumber(number)
  val result = Reads.of[ZonedDateTime].reads(jsNumber)
```

It should return a result quickly.

### Actual Behavior

It is CPU bound for an unknown period of time (I have never waited long enough for it to complete).

OlegYch:
# Pull Request Checklist

## Fixes

Fixes #180 

## Purpose

Rejects inexact numbers as input for date parsers, they are limited to Long anyway

## Background Context

`BigDecimal.toLong` takes inappropriate amount of time for something like `1000000000e1000000000`

### DIFF
--- a/play-json/jvm/src/test/scala/play/api/libs/json/ReadsSpec.scala
+++ b/play-json/jvm/src/test/scala/play/api/libs/json/ReadsSpec.scala
@@ -57,6 +57,12 @@ class ReadsSpec extends org.specs2.mutable.Specification {
         ))
     }
 
+    "not be read from invalid number" in {
+      reads(JsNumber(BigDecimal("1000000000e1000000000"))) aka "read date" must beLike {
+        case JsError((_, JsonValidationError("error.expected.long" :: Nil) :: Nil) :: Nil) => ok
+      }
+    }
+
     "not be read from invalid string" in {
       reads(JsString("invalid")) aka "read date" must beLike {
         case JsError((_, JsonValidationError(
@@ -224,6 +230,12 @@ class ReadsSpec extends org.specs2.mutable.Specification {
       }
     }
 
+    "not be read from invalid number" in {
+      reads(JsNumber(BigDecimal("1000000000e1000000000"))) aka "read date" must beLike {
+        case JsError((_, JsonValidationError("error.expected.long" :: Nil) :: Nil) :: Nil) => ok
+      }
+    }
+
     "be successfully read with default implicit" >> {
       "from '2011-12-03T10:15:30+01:00' (with TZ offset)" in {
         reads(JsString("2011-12-03T10:15:30+01:00")) aka "read date" must_== (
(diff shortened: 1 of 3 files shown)


OlegYch:
@dwijnand should i submit this against 2.6 branch too?

dwijnand:
I'm not sure, but I would think we probably want this fixed in 2.6 too. Whether or not that means a second PR.. I also don't know. 🙂

marcospereira:
Hi @OlegYch and @dwijnand,

Yes, we probably want to backport this to 2.6.x, but you won't need a new PR. We can cherry-pick on our side.

play-json/jvm/src/test/scala/play/api/libs/json/ReadsSpec.scala
```diff
@@ -224,6 +224,12 @@ class ReadsSpec extends org.specs2.mutable.Specification {
       }
     }
 
+    "not be read from invalid number" in {
```
marcospereira: Can we also have tests for `java.util.Date`, `java.sql.Date` and other affected?

play-json/shared/src/test/scala/play/api/libs/json/JsonSharedSpec.scala
```diff
@@ -270,7 +270,7 @@ class JsonSharedSpec extends WordSpec
         "key3" -> js.arr(1, "tutu")
       )
 
-      js.prettyPrint(jo) mustEqual ("""{
+      js.prettyPrint(jo).replaceAllLiterally("\r\n", "\n") mustEqual ("""{
```
marcospereira: Why is this necessary now?

OlegYch: otherwise the test breaks on windows

play-json/jvm/src/test/scala/play/api/libs/json/ReadsSpec.scala
OlegYch: added more tests

play-json/shared/src/test/scala/play/api/libs/json/JsonSharedSpec.scala
marcospereira: Oh, okay. Thanks for fixing that too. :-)

PR #184, opened 2018-09-24
Add `forall` method to `JsResult` with `Option#forall` semantics

aaabramov:
# Pull Request Checklist

## Purpose

Adds `forall` method to `JsResult` with `Option#forall` semantics

## References

Initial discussion: https://github.com/playframework/play-json/pull/178

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsResult.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsResult.scala
@@ -24,7 +24,7 @@ case class JsSuccess[T](value: T, path: JsPath = JsPath()) extends JsResult[T] {
   def contains[AA >: T](elem: AA): Boolean = elem == value
 
   def exists(p: T => Boolean): Boolean = p(value)
-  
+
   def forall(p: T => Boolean): Boolean = p(value)
 
   def repath(path: JsPath): JsResult[T] = JsSuccess(value, path ++ this.path)
@@ -68,11 +68,11 @@ case class JsError(errors: collection.Seq[(JsPath, collection.Seq[JsonValidation
   def flatMap[U](f: Nothing => JsResult[U]): JsResult[U] = this
 
   def foreach(f: Nothing => Unit): Unit = ()
-  
+
   def contains[AA >: Nothing](elem: AA): Boolean = false
 
   def exists(p: Nothing => Boolean): Boolean = false
-  
+
   def forall(p: Nothing => Boolean): Boolean = true
 
   def repath(path: JsPath): JsResult[Nothing] =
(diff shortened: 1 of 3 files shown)


play-json/shared/src/main/scala/play/api/libs/json/JsResult.scala
```diff
@@ -221,6 +225,12 @@ sealed trait JsResult[+A] { self =>
   /** If this result is successful than check value with predicate '''p''', otherwise return '''false''' */
   def exists(p: A => Boolean): Boolean
 
+  /**
+   * If this result is successful than check value with predicate '''p''', otherwise return '''true'''.
+   * Follows [[Option.forall]] semantics
```
cchantep: Option ~> Traversable

play-json/shared/src/test/scala/play/api/libs/json/JsResultSpec.scala
```diff
@@ -77,4 +77,24 @@ class JsResultSpec extends WordSpec with MustMatchers {
 
   }
 
+  "JsSuccess#forall" should {
+
+    "return true for JsSuccess(x * 2).forall(_ % 2 == 0)" in {
+      JsSuccess(2).forall(_ % 2 == 0) mustBe true
```
cchantep: `mustBe <boolean>` vs `assert` in previous PR ... coherency

aaabramov:
@cchantep PR updated.

play-json/shared/src/main/scala/play/api/libs/json/JsResult.scala
```diff
@@ -71,6 +73,8 @@ case class JsError(errors: collection.Seq[(JsPath, collection.Seq[JsonValidation
 
   def exists(p: Nothing => Boolean): Boolean = false
 
+  def forall(p: Nothing => Boolean): Boolean = true
+
```
renatocaval: I guess we must have it returning `true` just to align with other `forall` methods in Scala. However, this is something that always annoyed me. 

`forall` on the empty or error side always returns true in other libraries.

```scala
@ val optEmpty: Option[String] = None
optEmpty: Option[String] = None

@ optEmpty.forall(_.startsWith("a"))
res6: Boolean = true

@ optEmpty.forall(_.startsWith("b"))
res6: Boolean = true
```

In the above example, how does it even make any sense? If the option is empty, all predicates making assertions about its content should return false. The same is valid for Scala collections. 

We will have the same situation here.

renatocaval: I asked about it in the Scala forum the motivations for this behaviour. 
https://users.scala-lang.org/t/forall-semantics/3303

Not that I believe we should change it. We should keep it consistent with other libraries.

aaabramov: @renatocaval I realized that `_.forall` behavior is pretty lear after reading this blogpost: http://blog.originate.com/blog/2014/06/15/idiomatic-scala-your-options-do-not-match/

Specifically:
image
image

>It makes perfect sense in plain English: to say “all numbers are prime” is equivalent to saying “there does not exist a number that is not prime”.

renatocaval: Thanks @aaabramov. 
I suspected that there was a solid reason for that. TIL

PR #188, opened 2018-10-14
Reads & Writes for java.net.URI

cchantep:


build.sbt
```diff
@@ -120,7 +120,16 @@ lazy val `play-json` = crossProject(JVMPlatform, JSPlatform).crossType(CrossType
   .settings(commonSettings)
   .settings(playJsonMimaSettings)
   .settings(
-    mimaBinaryIssueFilters ++= Seq(
+  mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[ReversedMissingMethodProblem]("play.api.libs.json.LowPriorityDefaultReads.play$api$libs$json$LowPriorityDefaultReads$_setter_$uriReads_="),
```
cchantep: New instances

play-json/shared/src/main/scala/play/api/libs/json/Reads.scala
```diff
@@ -187,49 +192,10 @@ object Reads extends ConstraintReads with PathReads with DefaultReads with Gener
   implicit val JsArrayReducer = Reducer[JsValue, JsArray](js => JsArray(Array(js)))
 }
 
-/**
```
cchantep: Moved down

play-json/shared/src/main/scala/play/api/libs/json/Reads.scala
```diff
@@ -540,31 +542,55 @@ trait DefaultReads extends LowPriorityDefaultReads {
   }
 
   /**
-   * Deserializer for java.util.UUID
+   * Deserializer for java.net.URL
    */
-  class UUIDReader(checkValidity: Boolean) extends Reads[java.util.UUID] {
-    import java.util.UUID
+  implicit val urlReads: Reads[URL] = Reads[URL] {
```
cchantep: Instances for java.net

cchantep:
@dwijnand 

> I don't think java.net.URL should have Reads and Writes instances because of URL's problematic equals (it side-effects to check hosts resolve to the same IP!!).
> 
> So I think we should drop those.

Having `URL` as field can be found in multiple projects, so even if `URL` type has drawback (outside of Play-JSON scope), I do think it deserves to be handle properly there, or people will go on handling it with custom format on each project.

dwijnand:
I'd be fine to include it in play-json but outside of Reads/Writes implicit scope (i.e. not through the companion object).  So if we move it to another object that is imported, that's ok.  Something like `import play.json.UrlInstances._`.

plokhotnyuk:
Please take in account that `java.net.URL` has side-effecting implementation like DNS resolving.

So that can be [a source of some distributed attacks targeted to DNS](https://cira.ca/anatomy-ddos-attack-against-dns) services or just be [a source of unexpected delay in services which instantiate them](https://netbeans.org/bugzilla/show_bug.cgi?id=267962).

dwijnand:
@plokhotnyuk if you read up that's exactly my concern:

> I don't think java.net.URL should have Reads and Writes instances because of URL's problematic equals (it side-effects to check hosts resolve to the same IP!!).

cchantep:
Updated

dwijnand:
Thanks, @cchantep!

PR #191, opened 2018-10-16
Make it possible to configure the parsing for large big decimals

### LINKED ISSUE #187: Denial of service when parsing a JSON object with an unexpected field that has a big number

2.7.0-M1

Scala 2.12.7

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)

Ubuntu 16.04

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

Oracle JDK 11

### Library Dependencies

none

### Expected Behavior

_Sub-linear_ decreasing of throughput when length of the JSON object is increasing

### Actual Behavior

_Sub-quadratic_ decreasing of throughput when length of the JSON object is increasing

On contemporary CPUs parsing of such JSON object with an additional field that has of 1000000 decimal digits (~1Mb) can took more than 13 seconds:

```
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                                     (size)   Mode  Cnt         Score        Error  Units
[info] ExtractFieldsBenchmark.readAVSystemGenCodec        1  thrpt    5   3415309.675 ± 358732.424  ops/s
[info] ExtractFieldsBenchmark.readAVSystemGenCodec       10  thrpt    5   3429891.871 ± 135960.100  ops/s
[info] ExtractFieldsBenchmark.readAVSystemGenCodec      100  thrpt    5    686235.263 ± 832769.728  ops/s
[info] ExtractFieldsBenchmark.readAVSystemGenCodec     1000  thrpt    5    194588.852 ±   8165.939  ops/s
[info] ExtractFieldsBenchmark.readAVSystemGenCodec    10000  thrpt    5     28335.193 ±    911.581  ops/s
[info] ExtractFieldsBenchmark.readAVSystemGenCodec   100000  thrpt    5      2948.038 ±    128.163  ops/s
[info] ExtractFieldsBenchmark.readAVSystemGenCodec  1000000  thrpt    5       649.088 ±    199.346  ops/s
[info] ExtractFieldsBenchmark.readCirce                   1  thrpt    5   1181495.148 ± 302987.993  ops/s
[info] ExtractFieldsBenchmark.readCirce                  10  thrpt    5   1277915.025 ± 179880.016  ops/s
[info] ExtractFieldsBenchmark.readCirce                 100  thrpt    5   1277950.564 ± 256709.663  ops/s
[info] ExtractFieldsBenchmark.readCirce                1000  thrpt    5    816515.741 ±  15529.900  ops/s
[info] ExtractFieldsBenchmark.readCirce               10000  thrpt    5    146038.446 ±   2585.134  ops/s
[info] ExtractFieldsBenchmark.readCirce              100000  thrpt    5     16825.855 ±    468.669  ops/s
[info] ExtractFieldsBenchmark.readCirce             1000000  thrpt    5      1693.840 ±     59.649  ops/s
[info] ExtractFieldsBenchmark.readDslJsonJava             1  thrpt    5  11703558.471 ± 196574.764  ops/s
[info] ExtractFieldsBenchmark.readDslJsonJava            10  thrpt    5  10418348.204 ± 125349.933  ops/s
[info] ExtractFieldsBenchmark.readDslJsonJava           100  thrpt    5   4854474.847 ± 335999.431  ops/s
[info] ExtractFieldsBenchmark.readDslJsonJava          1000  thrpt    5    833174.664 ±  22787.464  ops/s
[info] ExtractFieldsBenchmark.readDslJsonJava         10000  thrpt    5     88047.329 ±    894.533  ops/s
[info] ExtractFieldsBenchmark.readDslJsonJava        100000  thrpt    5      9037.421 ±     97.407  ops/s
[info] ExtractFieldsBenchmark.readDslJsonJava       1000000  thrpt    5       918.420 ±     13.473  ops/s
[info] ExtractFieldsBenchmark.readJacksonScala            1  thrpt    5   2533509.752 ±  75854.375  ops/s
[info] ExtractFieldsBenchmark.readJacksonScala           10  thrpt    5   2521299.318 ±  37344.857  ops/s
[info] ExtractFieldsBenchmark.readJacksonScala          100  thrpt    5    868736.640 ±  19590.367  ops/s
[info] ExtractFieldsBenchmark.readJacksonScala         1000  thrpt    5     54637.764 ±   1922.976  ops/s
[info] ExtractFieldsBenchmark.readJacksonScala        10000  thrpt    5       723.644 ±     14.444  ops/s
[info] ExtractFieldsBenchmark.readJacksonScala       100000  thrpt    5         7.254 ±      0.414  ops/s
[info] ExtractFieldsBenchmark.readJacksonScala      1000000  thrpt    5         0.077 ±      0.001  ops/s
[info] ExtractFieldsBenchmark.readJsoniterScala           1  thrpt    5  17357927.186 ± 105168.663  ops/s
[info] ExtractFieldsBenchmark.readJsoniterScala          10  thrpt    5  15509884.192 ± 599007.176  ops/s
[info] ExtractFieldsBenchmark.readJsoniterScala         100  thrpt    5  10557719.687 ±  82797.425  ops/s
[info] ExtractFieldsBenchmark.readJsoniterScala        1000  thrpt    5   2306588.382 ±  15014.663  ops/s
[info] ExtractFieldsBenchmark.readJsoniterScala       10000  thrpt    5    252999.473 ±   2013.190  ops/s
[info] ExtractFieldsBenchmark.readJsoniterScala      100000  thrpt    5     24022.123 ±    490.780  ops/s
[info] ExtractFieldsBenchmark.readJsoniterScala     1000000  thrpt    5      2042.339 ±    118.757  ops/s
[info] ExtractFieldsBenchmark.readPlayJson                1  thrpt    5    928062.700 ±  35964.755  ops/s
[info] ExtractFieldsBenchmark.readPlayJson               10  thrpt    5    908324.771 ±  41278.052  ops/s
[info] ExtractFieldsBenchmark.readPlayJson              100  thrpt    5    538588.245 ±  58035.196  ops/s
[info] ExtractFieldsBenchmark.readPlayJson             1000  thrpt    5     52739.058 ±   5124.015  ops/s
[info] ExtractFieldsBenchmark.readPlayJson            10000  thrpt    5       743.426 ±      6.226  ops/s
[info] ExtractFieldsBenchmark.readPlayJson           100000  thrpt    5         7.351 ±      0.030  ops/s
[info] ExtractFieldsBenchmark.readPlayJson          1000000  thrpt    5         0.067 ±      0.018  ops/s
[info] ExtractFieldsBenchmark.readUPickle                 1  thrpt    5   3340922.046 ± 246892.139  ops/s
[info] ExtractFieldsBenchmark.readUPickle                10  thrpt    5   3483490.433 ±  39971.435  ops/s
[info] ExtractFieldsBenchmark.readUPickle               100  thrpt    5   2494567.445 ±  71404.382  ops/s
[info] ExtractFieldsBenchmark.readUPickle              1000  thrpt    5    814753.180 ±  30787.779  ops/s
[info] ExtractFieldsBenchmark.readUPickle             10000  thrpt    5    101384.553 ±   1049.347  ops/s
[info] ExtractFieldsBenchmark.readUPickle            100000  thrpt    5     10380.287 ±     43.464  ops/s
[info] ExtractFieldsBenchmark.readUPickle           1000000  thrpt    5       991.119 ±     60.797  ops/s
```

### Reproducible Test Case

To run that benchmarks on your JDK:
1. [Install latest version of `sbt`](https://www.scala-sbt.org/1.0/docs/Setup.html) and/or ensure that it already installed properly: 
```
sbt about
```

2. Clone `jsoniter-scala` repo: 
```
git clone https://github.com/plokhotnyuk/jsoniter-scala.git
```

3. Enter to the cloned directory and checkout for the specific branch:
```
cd jsoniter-scala
git checkout play-json-DoS-using-big-number
```

4. Run benchmarks using a path parameter to your JDK:
```
sbt -no-colors 'jsoniter-scala-benchmark/jmh:run -jvm /usr/lib/jvm/jdk-11/bin/java -wi 5 -i 5 .*ExtractFieldsBench.*'
```

marcospereira:
## Fixes

Fixes #187 

## Purpose

Parsing large big decimals (thing tens of hundred digits) and operating on these numbers
can be very CPU demanding. While play-json currently supports handling large numbers, it
is not practical on real-world applications and can expose them to DoS of service attacks.

This changes the way parsing happens to limit the size of such numbers based on
MathContext.DECIMAL128.

### DIFF
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -34,7 +34,7 @@ import scala.collection.mutable.{ ArrayBuffer, ListBuffer }
  *   val jsValue = mapper.readValue("""{"foo":"bar"}""", classOf[JsValue])
  * }}}
  */
-class PlayJsonModule(parserSettings: JsonParserSettings) extends SimpleModule("PlayJson", Version.unknownVersion()) {
+final class PlayJsonModule(parserSettings: JsonParserSettings) extends SimpleModule("PlayJson", Version.unknownVersion()) {
   override def setupModule(context: SetupContext): Unit = {
     context.addDeserializers(new PlayDeserializers(parserSettings))
     context.addSerializers(new PlaySerializers(parserSettings))
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -41,7 +41,7 @@ sealed class PlayJsonModule private[jackson] (parserSettings: JsonParserSettings
   }
 }
 
-@deprecated("Use PlayJsonModule class instead")
+@deprecated("Use PlayJsonModule class instead", "2.6.11")
 object PlayJsonModule extends PlayJsonModule(JsonParserSettings())
 
 // -- Serializers.
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -41,7 +41,8 @@ class PlayJsonModule(parserSettings: JsonParserSettings) extends SimpleModule("P
   }
 }
 
-object PlayJsonModule extends PlayJsonModule(JsonParserSettings.settings)
+@deprecated("Use PlayJsonModule class instead")
+object PlayJsonModule extends PlayJsonModule(JsonParserSettings())
 
 // -- Serializers.
 
@@ -246,7 +247,7 @@ private[jackson] class PlaySerializers extends Serializers.Base {
 
 private[json] object JacksonJson {
 
-  private lazy val mapper = (new ObjectMapper).registerModule(PlayJsonModule)
+  private lazy val mapper = (new ObjectMapper).registerModule(new PlayJsonModule(JsonParserSettings.settings))
 
   private lazy val jsonFactory = new JsonFactory(mapper)
 
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -5,7 +5,6 @@
 package play.api.libs.json.jackson
 
 import java.io.{ InputStream, StringWriter }
-import java.math.MathContext
 
 import com.fasterxml.jackson.core._
 import com.fasterxml.jackson.core.util.DefaultPrettyPrinter
@@ -30,16 +29,19 @@ import scala.collection.mutable.{ ArrayBuffer, ListBuffer }
  * {{{
  *   import com.fasterxml.jackson.databind.ObjectMapper
  *
- *   val mapper = new ObjectMapper().registerModule(PlayJsonModule)
+ *   val jsonParseSettings = JsonParserSettings()
+ *   val mapper = new ObjectMapper().registerModule(PlayJsonModule(jsonParseSettings))
  *   val jsValue = mapper.readValue("""{"foo":"bar"}""", classOf[JsValue])
  * }}}
  */
-object PlayJsonModule extends SimpleModule("PlayJson", Version.unknownVersion()) {
+class PlayJsonModule(parserSettings: JsonParserSettings) extends SimpleModule("PlayJson", Version.unknownVersion()) {
   override def setupModule(context: SetupContext): Unit = {
-    context.addDeserializers(new PlayDeserializers)
+    context.addDeserializers(new PlayDeserializers(parserSettings))
     context.addSerializers(new PlaySerializers)
   }
 }
+
+object PlayJsonModule extends PlayJsonModule(JsonParserSettings.settings)
 
 // -- Serializers.
 
@@ -121,7 +123,7 @@ private[jackson] case class ReadingMap(content: ListBuffer[(String, JsValue)]) e
 
 }
 
-private[jackson] class JsValueDeserializer(factory: TypeFactory, klass: Class[_]) extends JsonDeserializer[Object] {
+private[jackson] class JsValueDeserializer(factory: TypeFactory, klass: Class[_], parserSettings: JsonParserSettings) extends JsonDeserializer[Object] {
 
   override def isCachable: Boolean = true
 
@@ -141,15 +143,15 @@ private[jackson] class JsValueDeserializer(factory: TypeFactory, klass: Class[_]
     // There is a limit of how large the numbers can be since parsing extremely
     // large numbers (think thousand of digits) and operating on the parsed values
     // can potentially cause a DDoS.
-    if (inputLength > JacksonJson.BigDecimalLimits.DigitsLimit) {
+    if (inputLength > parserSettings.bigDecimalParseSettings.digitsLimit) {
       throw new IllegalArgumentException(s"""Number is larger than supported for field "${jp.currentName()}"""")
     }
 
     // Must create the BigDecimal with a MathContext that is consistent with the limits used.
-    val bigDecimal = BigDecimal(inputText, JacksonJson.BigDecimalLimits.DefaultMathContext)
+    val bigDecimal = BigDecimal(inputText, parserSettings.bigDecimalParseSettings.mathContext)
 
     // We should also avoid numbers with scale that are out of a safe limit
-    if (Math.abs(bigDecimal.scale) > JacksonJson.BigDecimalLimits.ScaleLimit) {
+    if (Math.abs(bigDecimal.scale) > parserSettings.bigDecimalParseSettings.scaleLimit) {
       throw new IllegalArgumentException(s"""Number scale (${bigDecimal.scale}) is out of limits for field "${jp.currentName()}"""")
     }
 
@@ -222,11 +224,11 @@ private[jackson] class JsValueDeserializer(factory: TypeFactory, klass: Class[_]
   override val getNullValue = JsNull
 }
 
-private[jackson] class PlayDeserializers extends Deserializers.Base {
+private[jackson] class PlayDeserializers(parserSettings: JsonParserSettings) extends Deserializers.Base {
   override def findBeanDeserializer(javaType: JavaType, config: DeserializationConfig, beanDesc: BeanDescription) = {
     val klass = javaType.getRawClass
     if (classOf[JsValue].isAssignableFrom(klass) || klass == JsNull.getClass) {
-      new JsValueDeserializer(config.getTypeFactory, klass)
+      new JsValueDeserializer(config.getTypeFactory, klass, parserSettings)
     } else null
   }
 }
@@ -244,28 +246,9 @@ private[jackson] class PlaySerializers extends Serializers.Base {
 
 private[json] object JacksonJson {
 
-  /**
-   * Define limits for parsing BigDecimal numbers.
-   *
-   * By default, we are using MathContext.DECIMAL128 and then the limits are define in its terms.
-   */
-  object BigDecimalLimits {
+  private lazy val mapper = (new ObjectMapper).registerModule(PlayJsonModule)
 
-    val DefaultMathContext: MathContext = MathContext.DECIMAL128
-
-    // Limit for the scale considering the MathContext of 128
-    // limit for scale for decimal128: BigDecimal("0." + "0" * 33 + "1e-6143", java.math.MathContext.DECIMAL128).scale + 1
-    val ScaleLimit: Int = 6178
-
-    // 307 digits should be the correct value for 128 bytes. But we are using 310
-    // because Play JSON uses BigDecimal to parse any number including Doubles and
-    // Doubles max value has 309 digits, so we are using 310 here
-    val DigitsLimit: Int = 310
-  }
-
-  private val mapper = (new ObjectMapper).registerModule(PlayJsonModule)
-
-  private val jsonFactory = new JsonFactory(mapper)
+  private lazy val jsonFactory = new JsonFactory(mapper)
 
   private def stringJsonGenerator(out: java.io.StringWriter) =
     jsonFactory.createGenerator(out)
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -34,7 +34,7 @@ import scala.collection.mutable.{ ArrayBuffer, ListBuffer }
  *   val jsValue = mapper.readValue("""{"foo":"bar"}""", classOf[JsValue])
  * }}}
  */
-final class PlayJsonModule(parserSettings: JsonParserSettings) extends SimpleModule("PlayJson", Version.unknownVersion()) {
+sealed class PlayJsonModule private[jackson] (parserSettings: JsonParserSettings) extends SimpleModule("PlayJson", Version.unknownVersion()) {
   override def setupModule(context: SetupContext): Unit = {
     context.addDeserializers(new PlayDeserializers(parserSettings))
     context.addSerializers(new PlaySerializers(parserSettings))
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -37,7 +37,7 @@ import scala.collection.mutable.{ ArrayBuffer, ListBuffer }
 class PlayJsonModule(parserSettings: JsonParserSettings) extends SimpleModule("PlayJson", Version.unknownVersion()) {
   override def setupModule(context: SetupContext): Unit = {
     context.addDeserializers(new PlayDeserializers(parserSettings))
-    context.addSerializers(new PlaySerializers)
+    context.addSerializers(new PlaySerializers(parserSettings))
   }
 }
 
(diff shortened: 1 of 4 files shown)


play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -141,7 +142,23 @@ private[jackson] class JsValueDeserializer(factory: TypeFactory, klass: Class[_]
...
+        val inputText = jp.getText
+        val inputLength = inputText.length
+
+        // There is a limit of how large the numbers can be since parsing extremely
+        // large numbers (think thousand of digits) and operating on the parsed values
+        // can potentially cause a DDoS.
+        if (inputLength > JacksonJson.BigDecimalLimits.DigitsLimit) throw new IllegalArgumentException("Number is larger than supported")
+
+        // Must create the BigDecimal with a MathContext that is consistent with the limits used.
+        val bigDecimal = BigDecimal(inputText, JacksonJson.BigDecimalLimits.DefaultMathContext)
```
marcospereira: This is one of the crucial changes: using `MathContext.DECIMAL128` imposes a precision of 34 and then numbers requiring a higher precision will be truncated. But, a precision of 34 is good enough for most real cases.

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -60,66 +83,369 @@ class JsonSpec extends org.specs2.mutable.Specification {
...
-    }
-
-    "with default/lenient date format without millis or time zone" in {
-      val postJson =
-        """{"body": "foobar", "created_at": "2011-04-22T13:33:48"}"""
-      val expectedPost = Post("foobar", Some(postDateWithTZ(TimeZone.getDefault)))
-
-      Json.parse(postJson).as[Post](LenientPostFormat).
-        aka("parsed") must_== expectedPost
+    "when parsing numbers" in {
```
marcospereira: There are some isolated tests for parsing numbers but decided to add these here since the changes made have an effect and number types.

These tests are also closer to real usage of play-json.

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -60,66 +83,369 @@ class JsonSpec extends org.specs2.mutable.Specification {
...
+
+      "for BigDecimals" should {
+        "maintain precision when parsing BigDecimals within precision limit" in {
+          val n = BigDecimal("12345678901234567890.123456789")
+          val json = toJson(n)
+          parse(stringify(json)) mustEqual json
+        }
+
+        "truncate when exceeding the precision limit" in {
+          // lsat two six are exceeding 34 precision limit
```
marcospereira:
```suggestion
          // last two six are exceeding 34 precision limit
```

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -60,66 +83,369 @@ class JsonSpec extends org.specs2.mutable.Specification {
...
+      "for BigDecimals" should {
+        "maintain precision when parsing BigDecimals within precision limit" in {
+          val n = BigDecimal("12345678901234567890.123456789")
+          val json = toJson(n)
+          parse(stringify(json)) mustEqual json
+        }
+
+        "truncate when exceeding the precision limit" in {
+          // last two six are exceeding 34 precision limit
+          val n = BigDecimal("10.12345678912345678912345678912345666")
```
ignasi35: Use `0` on the `123456789` so it's easier to count.

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -60,66 +83,369 @@ class JsonSpec extends org.specs2.mutable.Specification {
...
+               |{
+               |  "float": 1,
+               |  "double": $outOfLimits
+               |}""".stripMargin)
+            .as[FloatNumbers].double mustEqual Double.MinValue
+        }
+      }
+
+      "for BigDecimals" should {
+        "maintain precision when parsing BigDecimals within precision limit" in {
```
ignasi35: Can we add a note with breadcrumbs detailing where precision is set: 
```scala
  // note: precision refers to `JacksonJson.BigDecimalLimits.DefaultMathContext.getPrecision`
```
?

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -60,66 +83,369 @@ class JsonSpec extends org.specs2.mutable.Specification {
...
-    "with default/lenient date format without millis or time zone" in {
-      val postJson =
-        """{"body": "foobar", "created_at": "2011-04-22T13:33:48"}"""
-      val expectedPost = Post("foobar", Some(postDateWithTZ(TimeZone.getDefault)))
-
-      Json.parse(postJson).as[Post](LenientPostFormat).
-        aka("parsed") must_== expectedPost
+    "when parsing numbers" in {
+      "for Long" should {
+
```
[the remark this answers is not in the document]
marcospereira: Done.

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
marcospereira: Done. :-)

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -141,7 +142,23 @@ private[jackson] class JsValueDeserializer(factory: TypeFactory, klass: Class[_]
...
+
+        // We should also avoid numbers with scale that are out of a safe limit
+        if (Math.abs(bigDecimal.scale) > JacksonJson.BigDecimalLimits.ScaleLimit) throw new IllegalArgumentException(s"Number scale (${bigDecimal.scale}) is out of limits")
+
+        (Some(JsNumber(bigDecimal)), parserContext)
+      }
```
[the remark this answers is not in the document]
marcospereira: > (nitpick) a 15-line block breaks readability of the switch statement.

Solved.

> (nitpick) how does JacksonJson.BigDecimalLimits relate to serialiser limits? We are destructive in serilisation and relaxed in parsing (still accept huge values). The inconsistency already existed so we may want to improve in a separate PR.

Well, the serializer is not imposing any limits besides how to represent the number (using plain or scientific notation). My rationale here was that it is up to the user to validate if the number is within limits when serializing (considering large numbers) because the BigDecimal is already there, it already exists and was computed before serializing.

WDYT?

> (nitpick) IT'd be great if the exceptions included the name of the field containing the offending data. Something like val maybeFieldName = parserContext.find(_.isInstanceOf[KeyRead]).map{_.asInstanceOf[KeyRead].fieldName}? This is a security validation/restriction so I'm not 100% sure it's really necessary to add this extra info though.

Done.

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
marcospereira: Done.

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -221,6 +244,25 @@ private[jackson] class PlaySerializers extends Serializers.Base {
 
 private[json] object JacksonJson {
 
+  /**
+   * Define limits for parsing BigDecimal numbers.
+   *
+   * By default, we are using MathContext.DECIMAL128 and then the limits are define in its terms.
+   */
+  object BigDecimalLimits {
```
plokhotnyuk: It is great to have some safe defaults, but shouldn't also be an ability to configure them for different cases?

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -30,21 +29,25 @@ import scala.collection.mutable.{ ArrayBuffer, ListBuffer }
...
-object PlayJsonModule extends SimpleModule("PlayJson", Version.unknownVersion()) {
+class PlayJsonModule(parserSettings: JsonParserSettings) extends SimpleModule("PlayJson", Version.unknownVersion()) {
   override def setupModule(context: SetupContext): Unit = {
-    context.addDeserializers(new PlayDeserializers)
+    context.addDeserializers(new PlayDeserializers(parserSettings))
     context.addSerializers(new PlaySerializers)
   }
 }
 
+object PlayJsonModule extends PlayJsonModule(JsonParserSettings.settings)
```
dwijnand: Having to race against anything else on the classpath initializing this and therefore locking it to whatever it's currently set to feels wrong.

I propose the follow:

* deprecate public `PlayJsonModule` (the object)
* change it to use `JsonParserSettings()`, the (safe) default, so it's behavior is stable
* change `JacksonJson` (which is `package[json]`) from object to class so we don't introduce a race there either (you can then revert back the lazy modifiers)

marcospereira: About the `JacksonJson`, what I mean was:

```scala
private lazy val jackson = JacksonJson(JsonParseSettings.settings)
```

In the end, maybe what we need is a way to initialize `mapper` and `jsonFactory` in `JacksonJson`.

dwijnand: No, personally I wouldn't have any lazy vals in objects and just create `JacksonJson` every time you need to parse.

marcospereira: > create JacksonJson every time you need to parse.

Which implies that we are creating `ObjectMapper` every time also and that is not a good idea (it is expensive to created). As far as I know, `ObjectMapper` is thread-safe and meant to be re-used across the lifetime of the application.

dwijnand: Then I guess make `EnvReads` and `EnvWrites` classes and deprecate the objects?

dwijnand: Actually, why do we need to mutate object `JsonParserSettings`?  Can't we just wire a different `JsonParserSettings`?

dwijnand: @marcospereira If you want to get this in for RC1, I suggest leaving out the global mutable state part.

marcospereira: > Actually, why do we need to mutate object JsonParserSettings? Can't we just wire a different JsonParserSettings instance?

That is why I was suggesting to instead have a way to initialize and mutate `ObjectMapper` in `JacksonJson`. Looks more reasonable.

> @marcospereira If you want to get this in for RC1, I suggest leaving out the global mutable state part.

We can do RC1 without this change.

dwijnand: I mean a non-static way to wire it.

Basically, I would prefer not introducing more global mutable state.  So if a user wants to customise this behavior, they forgo these statically initialised object API, they create their own `JsonParserSettings` (or `JacksonJson`, `EnvReads`/`EnvWrites`) and goes about their business.  WDYT?

marcospereira: Yeah, but that implies more profound changes. The main entry point for play-json API is to call `Json.toJson(something)`, which uses global state (`ObjectMapper`, `JsonFactory`, etc.). Change this is way out of the scope for this PR.

What we need here is to find a way to:

1. Fix the big decimal parsing.
2. And configure the bounds for the big decimal parsing.

I agree that having mutable global state is not ideal, but how to achieve that right now without changing too much of play-json?

dwijnand:
Pulled the initial commits in #200

marcospereira:
@dwijnand made the classes final. Also add another commit to make `MaxPlain`/`MinPlain` configurable.

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -30,30 +29,28 @@ import scala.collection.mutable.{ ArrayBuffer, ListBuffer }
...
  *   import com.fasterxml.jackson.databind.ObjectMapper
  *
- *   val mapper = new ObjectMapper().registerModule(PlayJsonModule)
+ *   val jsonParseSettings = JsonParserSettings()
+ *   val mapper = new ObjectMapper().registerModule(PlayJsonModule(jsonParseSettings))
  *   val jsValue = mapper.readValue("""{"foo":"bar"}""", classOf[JsValue])
  * }}}
  */
-object PlayJsonModule extends SimpleModule("PlayJson", Version.unknownVersion()) {
+class PlayJsonModule(parserSettings: JsonParserSettings) extends SimpleModule("PlayJson", Version.unknownVersion()) {
```
dwijnand:
```suggestion
final class PlayJsonModule(parserSettings: JsonParserSettings) extends SimpleModule("PlayJson", Version.unknownVersion()) {
```

dwijnand:
```
[error]  * object play.api.libs.json.jackson.JsValueSerializer does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[MissingClassProblem]("play.api.libs.json.jackson.JsValueSerializer$")
[error]  * method this()Unit in class play.api.libs.json.jackson.PlaySerializers does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("play.api.libs.json.jackson.PlaySerializers.this")
````

**Comment by avdv** [comment]
Hi.

I am using the PlayJsonModule like this:
```scala
val mapper = new ObjectMapper(new YAMLFactory()).registerModule(PlayJsonModule)
```
and I am using `-Xfatal-warnings` in my project. Now, after upgrading to version 2.6.11 I get this ~warning~ error:
```
[error] /codegen/src/main/scala/ModelGenerator.scala:95:75: object PlayJsonModule in package jackson is deprecated (since 2.6.11): Use PlayJsonModule class instead
[error]   private val mapper = new ObjectMapper(new YAMLFactory()).registerModule(PlayJsonModule)
[error]                                                                           ^
[error] No warnings can be incurred under -Xfatal-warnings.
```
Since the `PlayJsonModule` class is `private[jackson]` I am stuck. :sos:

marcospereira:
Hi @avdv, 

It is okay to open up this constructor, I think. If you want to send a PR, I can review it and release a new version later.

avdv:
Hi @marcospereira,

> It is okay to open up this constructor, I think. If you want to send a PR, I can review it and release a new version later.

Done: https://github.com/playframework/play-json/pull/206

Should I also backport to the 2.6 branch?

PR #192, opened 2018-10-17
Widen Reads for compatible types

cchantep:
```scala
import play.api.libs.json.Reads

def simple(r: Reads[Dog]): Reads[Animal] = r.widen[Animal]

def combine(dog: Reads[Dog], cat: Reads[Cat]): Reads[Animal] =
  dog.orElse(cat).orElse(Reads.failed[Animal]("Unsupported Animal"))
```

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
@@ -16,7 +16,7 @@ import scala.reflect.ClassTag
 @implicitNotFound(
   "No Json serializer found for type ${A}. Try to implement an implicit Writes or Format for this type."
 )
-trait Writes[-A] {
+trait Writes[-A] { self =>
   /**
    * Converts the `A` value into a [[JsValue]].
    */
@@ -26,17 +26,21 @@ trait Writes[-A] {
    * Returns a new instance that first converts a `B` value to a `A` one,
    * before converting this `A` value into a [[JsValue]].
    */
-  def contramap[B](f: B => A): Writes[B] = Writes[B](b => this.writes(f(b)))
+  def contramap[B](f: B => A): Writes[B] = Writes[B](b => self.writes(f(b)))
 
   /**
    * Transforms the resulting [[JsValue]] using transformer function.
    */
-  def transform(transformer: JsValue => JsValue): Writes[A] = Writes[A] { a => transformer(this.writes(a)) }
+  def transform(transformer: JsValue => JsValue): Writes[A] = Writes[A] { a =>
+    transformer(self.writes(a))
+  }
 
   /**
    * Transforms the resulting [[JsValue]] using a `Writes[JsValue]`.
    */
-  def transform(transformer: Writes[JsValue]): Writes[A] = Writes[A] { a => transformer.writes(this.writes(a)) }
+  def transform(transformer: Writes[JsValue]): Writes[A] = Writes[A] { a =>
+    transformer.writes(self.writes(a))
+  }
 }
 
 @implicitNotFound(
--- a/play-json/shared/src/test/scala/play/api/libs/json/ReadsSharedSpec.scala
+++ b/play-json/shared/src/test/scala/play/api/libs/json/ReadsSharedSpec.scala
@@ -11,8 +11,8 @@ import java.util.Locale
 import org.scalatest._
 
 class ReadsSharedSpec extends WordSpec with MustMatchers {
-  "Reads flatMap" should {
-    "not repath the second result" when {
+  "Reads" should {
+    "not repath the second result on flatMap" when {
       val aPath = JsPath \ "a"
       val readsA: Reads[String] = aPath.read[String]
       val value = "string"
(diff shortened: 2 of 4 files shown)


play-json/shared/src/test/scala/play/api/libs/json/ReadsSharedSpec.scala
```diff
@@ -33,6 +33,22 @@ class ReadsSharedSpec extends WordSpec with MustMatchers {
           )))))
       }
     }
+
+    "widen" in {
+      // !! Keep type ascriptions
+      val orig: Reads[List[String]] = implicitly[Reads[List[String]]]
+      val widened: Reads[Traversable[String]] = orig.widen
```
cchantep: If just `= orig`:

```
[error]  found   : play.api.libs.json.Reads[List[String]]
[error]  required: play.api.libs.json.Reads[Traversable[String]]
[error] Note: List[String] <: Traversable[String], but trait Reads is invariant in type A.
[error] You may wish to define A as +A instead. (SLS 4.5)
[error]       val widened: Reads[Traversable[String]] = orig
```

dwijnand:
This LGTM, but isn't it a binary-incompatible change for Scala 2.11?

cchantep:
It adds new function without altering existing one, which has already been done since last release (in fact other breaking change as already been applied that makes it not fully binary compatible but API compatible).

dwijnand:
I would assume that binary compatibility for a library like play-json is something that should be upheld...

dwijnand:
@cchantep given this is a utility function, why don't you implement it as an enrichment method for `Reads`?  Avoids the whole binary compatibility concern.

play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
```diff
@@ -26,17 +26,21 @@ trait Writes[-A] {
    * Returns a new instance that first converts a `B` value to a `A` one,
    * before converting this `A` value into a [[JsValue]].
    */
-  def contramap[B](f: B => A): Writes[B] = Writes[B](b => this.writes(f(b)))
+  def contramap[B](f: B => A): Writes[B] = Writes[B](b => self.writes(f(b)))
```
dwijnand: Just out of interest, what's the reason for this set of changes?

cchantep: Use `self` as elsewhere in the class

cchantep:
Adding a similar function anywhere (e.g. in the companion object) would result in a similar MiMa warning, plus it won't be coherent with the way other combinators are provided.

dwijnand:
Why would introducing an implicit class in the companion object result in a MiMa warning?  I agree it would be inconsistent, but such is the cost of compatibility.

play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
dwijnand: I guess you mean this makes it consistent with the use of `self` in the `Reads` class.  👌

cchantep:
MiMa would detect it as present in new version but not in the old one, warning for the object rather than the trait but it won't make the overall lib more bincompat.

cchantep:
FYI there were a recent similar discussion about compat https://github.com/playframework/play-json/pull/158#discussion_r188506774

dwijnand:
Introducing a class (and an implicit conversion to that class) isn't binary incompatible, though.

dwijnand:
When I get some time I'll demonstrate what I mean, maybe we're just misunderstanding each other! 🙂

cchantep:
Not check the MiMa, bu anyway I'm not for introducing implicit automagic/pimp my lib for such purpose.

dwijnand:
The only other choices, in the face of binary compatibility and Scala 2.11, I see, are:

* wait until Scala 2.11 support is dropped; or
* make it a function on the `Reads` object instead of on the `Reads` trait

That last one would be that instead of

```scala
def simple(r: Reads[Dog]): Reads[Animal] = r.widen[Animal]
```

it would be

```scala
def simple(r: Reads[Dog]): Reads[Animal] = Reads.widen[Animal](r)
```

dwijnand:
Personally, I think we should go with implementing this as an extension method, already in implicit scope.

dwijnand:
Ah yes, thank you @cchantep for reminding me again.  You mentioned it earlier but I didn't take any action.  Is it the change in #158 that you're referring to?

Personally, I feel like binary compatibility is very important.  I'm sorry if I'm coming across as confrontational or intruding, it's not my intent.

Could you explain to me why you feel against my idea of using an implicit class, please, so that I can better understand?

Thank you.

cchantep:
I'm generally against using implicit conversions, which have bad impact on readability and maintenance of the code, specially for a public API.

I'm way more at ease to break bincompat with new function (which I personally don't really consider as an issue as long as the API/source compatibility is maintained, and which is anyway broken by several occurrences in the different Play lib).

If it comes to that, I'd not having new features (freeze) for 2.11, rather than implement it with "pimp-my-lib" with implicit conversion.

dwijnand:
Alright, then as soon as we drop 2.11 we can add this again!

cchantep:
I don't see why it's closed, whereas it can at least be updated/merged for 2.12, 2.13.

dwijnand:
My thinking is we can resubmit once we've dropped Scala 2.11.

Issue #185, opened 2018-10-08
2.12.7 regression: Json.obj implicits


aaabramov:
**Note: This issue is reproducible only with scala 2.12.7**.
Scala 2.12.6 works properly.

### Play JSON Version (2.5.x / etc)

**2.6.10**

### API (Scala / Java / Neither / Both)

Scala

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)

```
Darwin 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018
```

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

```
> java -version
java version "1.8.0_144"
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)

[project] $ scalaVersion
[info] 2.12.7
```

### Expected Behavior

The following code should be properly compiled:
```
val url = Option(new URL(""))
val json = Json.obj("url" -> url.map(_.toString).getOrElse("#"))
```

### Actual Behavior

```
[error] .../Example.scala:23: type mismatch;
[error]  found   : Object
[error]  required: play.api.libs.json.Json.JsValueWrapper
[error]     "url" -> url.map(_.toString).getOrElse("#")
[error]                                           ^
[error] one error found
[error] (compile:compileIncremental) Compilation failed
```

### Workaround:

```
val url = Option(new URL(""))
val json = Json.obj(
  "url" -> (url.map(_.toString).getOrElse("#"): String)
)
```

### Reproducible Test Case

```
// Scala 2.12.7 
val url = Option(new URL(""))
val json = Json.obj("url" -> url.map(_.toString).getOrElse("#"))
```

aaabramov:
@gmethvin @dwijnand would you like to take a loot at this?

dwijnand:
I can't reproduce it working with 2.12.6:

```scala
Welcome to Scala 2.12.6 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_144).
Type in expressions for evaluation. Or try :help.

scala> import play.api.libs.json._
import play.api.libs.json._

scala> val url = Option(new java.net.URL("http://a.b"))
url: Option[java.net.URL] = Some(http://a.b)

scala> val json = Json.obj("url" -> url.map(_.toString).getOrElse("#"))
<console>:15: error: type mismatch;
[...]

PR #201, opened 2018-11-14
Add a case using Json.Obj implicits

dwijnand:
## Purpose

This is a test case for #185


dwijnand:
Actually, as stated in https://github.com/playframework/play-json/issues/185#issuecomment-432680544, this doesn't seem to be a regression in 2.12.7 as it doesn't reproduce in 2.12.6.

PR #202, opened 2018-11-21
[2.6.x] Cherry pick JSON parsing

dwijnand:
These two commits are the cherry-pick of #200 and #191.

I want someone to review that the MiMa binary changes are ok for 2.6.x (because I suspect they're not).


dwijnand:
Should we be concerned about Scala 2.11 users of DefaultReads/DefaultWrites?

dwijnand:
Actually, no one consumes DefaultReads/Writes extending objects to call methods off of it, at most they mix it in and then import its implicits off it.  So no binary risk.

PR #222, opened 2019-01-16
[2.7.x]: Configure binary compatibility check

marcospereira:


dwijnand:
Fast-forwarded master to include this change too.

PR #230, opened 2019-01-25
Replace scala-collection-compat with ScalaCollectionCompat

dwijnand:


marcospereira:
Do we need a backport to 2.7.x here?

dwijnand:
Yep.

PR #231, opened 2019-01-26
Adds support for MapView

### LINKED ISSUE #225: Make MapView (2.13 only?) a JSON map
This is bug/usability problem in play-json where [MapView](https://scala-lang.org/files/archive/api/2.13.0-M5/scala/collection/MapView.html) is not serialized into a Json object like a Map, but instead into an Json Array, because it is MapView is an Iterable...

_Originally posted by @jtjeferreira in https://github.com/playframework/playframework/pull/8986_

jtjeferreira:
# Pull Request Checklist

* [X] Have you signed the [Typesafe CLA](https://www.typesafe.com/contribute/cla)?
* [X] Have you added copyright headers to new files?

## Fixes

Fixes #225 

## Purpose

Adds support for [MapView](https://scala-lang.org/files/archive/api/2.13.x/scala/collection/MapView.html)

## Background Context

Instead of duplicating the code of `Writes.scala` in a 2.12 and 2.13 folders I extracted the different writers into a MapWrites companion object...

## References

#225

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -134,6 +134,17 @@ lazy val `play-json` = crossProject(JVMPlatform, JSPlatform).crossType(CrossType
         case Some((2, 13)) => Seq()
         case _ => Seq(compilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full))
       }),
+    // Adds a `src/main/scala-2.13+` source directory for Scala 2.13 and newer
+    // and a `src/main/scala-2.13-` source directory for Scala version older than 2.13
+    unmanagedSourceDirectories in Compile ++= {
+      (unmanagedSourceDirectories in Compile).value.map { dir =>
+        val sv = scalaVersion.value
+        CrossVersion.partialVersion(sv) match {
+          case Some((2, 13)) => file(dir.getPath ++ "-2.13+")
+          case _             => file(dir.getPath ++ "-2.13-")
+        }
+      }
+    },
     sourceGenerators in Compile += Def.task{
       val dir = (sourceManaged in Compile).value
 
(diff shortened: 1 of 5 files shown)


jtjeferreira:
I was hoping that 1a07724 would fail in travis, but since I pushed 59cd24d right after, travis built that version... I will push force only the first commit, wait for travis and then push the 2nd commit...

jtjeferreira:
So as expected the test fails:

```
[info] Map Writes
[info] - should write lazy maps *** FAILED *** (90 milliseconds)
[info]   [["a",2]] did not equal {"a":2} (WritesSharedSpec.scala:57)
```

build.sbt
```diff
@@ -134,6 +134,17 @@ lazy val `play-json` = crossProject(JVMPlatform, JSPlatform).crossType(CrossType
         case Some((2, 13)) => Seq()
         case _ => Seq(compilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full))
       }),
+    // Adds a `src/main/scala-2.13+` source directory for Scala 2.13 and newer
```
marcospereira: This is conflicting with https://github.com/playframework/play-json/pull/230.

jtjeferreira: fixed!

dwijnand:
Being an implementation type, I'm not sure if it's the best idea to build off of `MapOps`.  I would've just added an implementation for 2.13's `MapView` and included it.

jtjeferreira:
> I would've just added an implementation for 2.13's `MapView` and included it.

I also tried that, but I didn't manage to do it without duplicating the whole `Writes.scala` file...

dwijnand:
You can just put it into a trait that you mix into `DefaultWrites` (or something like that), and the trait is an empty trait for <2.13, and defines an instance for `MapView` for >=2.13.

jtjeferreira:
> You can just put it into a trait that you mix into `DefaultWrites` (or something like that), and the trait is an empty trait for <2.13, and defines an instance for `MapView` for >=2.13.

I think I tried that but was hitting ambiguous implicit with traversableWrites... But I can give it another try...

dwijnand:
I can believe that to be true, with them both being traversable.

I'll ask the Scala Team if we should build off of `MapOps`.

PR #235, opened 2019-02-15
Add probot settings

marcospereira:
Add probot settings configuration.

## References

https://github.com/probot/settings

### DIFF
--- a/.github/settings.yml
+++ b/.github/settings.yml
@@ -1,10 +1,9 @@
 # These settings are synced to GitHub by https://probot.github.io/apps/settings/
 repository:
-  homepage: "https://www.playframework.com/documentation/latest/JavaWS"
   topics: playframework, json, dsl, convenience-macros, json-ast, automatic-conversion, scala, scalajs, serialization, serializer
   private: false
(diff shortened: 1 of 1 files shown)


.github/settings.yml
```diff
@@ -0,0 +1,119 @@
...
+teams:
+  - name: core
+    permission: admin
+  - name: integrators
+    permission: write
+  - name: write-bots
+    permission: write
+
+branches:
+  - name: master
```
renatocaval: These two rules are exactly the same, shouldn't we just use `*` for matching all branches?

marcospereira: Yep, this is something to discuss. The existing difference is that master branch is more relaxed and it is not requiring the status checks and also it is not requiring the branch to be updated before merging.

I keep it separated here so that we can decide if we keep master like what we have today or if we make it more strict like the stable branches.

marcospereira: And by the way, this is not consistent across our repositories. Some enforce the pull request status checks, some don't; some require the branch to be updated, some don't.

Issue #236, opened 2019-02-18
JsObject/JsValue does not preserve order of fields in version 2.6.11 and later


gyoho:
### Play JSON Version >= 2.6.11

JsObject seems not preserving the fields order when removing a field. This seems occurring in `2.6.11` and later version. As I went through the commit, I found that the underlying collection type has been switched from `LinkedHashMap` provided by Scala library to Java's `LinedHashMap`. I suspect that commit caused this issue. (https://github.com/playframework/play-json/pull/193/files)

### API
- Scala version `2.12.8`
- SBT version `1.2.8`

### JDK
```
java version "1.8.0_201"
Java(TM) SE Runtime Environment (build 1.8.0_201-b09)
Java HotSpot(TM) 64-Bit Server VM (build 25.201-b09, mixed mode)
```

### Test Case

```scala
import play.api.libs.json._

case class Test(foo: Int, bar: String, baz: Boolean)
val testFmt: Format[Test] = Json.format[Test]
val test = Test(1, "bar", true)
val jsObj = testFmt.writes(test).as[JsObject]

val outOfOrder = jsObj - "baz"
```

#### Expected Behavior
```
outOfOrder should be {"foo":1,"bar":"bar"}
```

#### Actual Behavior
```
outOfOrder actually is {"bar":"bar","foo":1}
```

I couldn't find a pattern of the order, but apparently this happens when there are more than two fields in JsObject.

levinson:
Good investigative work and thanks for opening the bug on play-json! I ran into this issue earlier when attempting to upgrade to a newer Play version and ended up backing out of the change.

exp0nge:
@dwijnand What's the state of this? This is a breaking change.

dwijnand:
Fixed in #253.

PR #253, opened 2019-04-06
Preserve order of fields in JsObject

tototoshi:
Fixed https://github.com/playframework/play-json/issues/236

By https://github.com/playframework/play-json/commit/4a3e3b036c0d79e3c01a066cdb929f520052e7ae, JsObject#underlying are changed from scala.collection.mutable.LinkedHashMap to
scala.collection.Map that wraps java.util.LinkedHashMap.
Since new instances created With ++/+/- methods of the map are not LinkedHashMap based,
JsObject.++/+/- don't preserve the order of fields. We need to use ++=/+=/-= instead.


ignasi35:
I'm a bit divided.

https://github.com/playframework/play-json/commit/4a3e3b036c0d79e3c01a066cdb929f520052e7ae added a performance hit for the sake of security and this change increases allocations and object shuffling for correctness.

I think we must merge but I wonder if we're causing a perf degradation.

renatocaval:
@ignasi35, 4a3e3b036c0d79e3c01a066cdb929f520052e7ae moves to `java.util.LinkedHashMap` to better handle hash colision. But the intention is defeated whenever the `JsObject` is manipulated because we fallback to `mutable.LinkedHashMap`. Without this fix, 4a3e3b036c0d79e3c01a066cdb929f520052e7ae is practically useless, if I understand it correctly. 

Maybe we can try to avoid the extra allocation by keeping a `java.util.LinkedHashMap` as the underlying type.

ignasi35:
> Without this fix, [4a3e3b0](https://github.com/playframework/play-json/commit/4a3e3b036c0d79e3c01a066cdb929f520052e7ae) is practically useless flawed, if I understand it correctly.

True, I missed that detail. But my point still holds.

exp0nge:
Possible to merge this and backport to 2.7.x?

marcospereira:
@exp0nge right now we don't have the bandwidth to backport it to 2.7.x. I appreciate the effort done here, but I think we should revert the changes here. According to [RFC 7159](https://tools.ietf.org/html/rfc7159#page-3), the fields in an object are unordered:

> An object is an **unordered** collection of zero or more name/value pairs, where a name is a string, and a value is a string, number, boolean, null, object, or array.

I don't see a reason to go against the RFC and ensure the order here, and this is not expressed in the API either. Why is the order relevant to you here?

sihil:
My 2p worth on this. 

The JsObject has always had a comment on `fields` saying that the order from the constructor is preserved which indicates that this was intentional, albeit untested, behaviour. It follows that further additions should also preserve order.

Whilst the spec doesn't specify any ordering, it is, for example, helpful for humans reading API responses to have fields logically grouped together and without this fix this would not be possible. Thanks @tototoshi.

PR #254, opened 2019-04-08
Scala 2.13.0-RC1 support

marcospereira:
Backport won't be clean since here we are removing Scala 2.11 and we should keep it in 2.7.x branch.


dwijnand:
2.7.x cb7da35
2.6.x https://github.com/playframework/play-json/pull/255

PR #262, opened 2019-04-19
Actually run MiMa in CI

dwijnand:
... and cache the sbt launcher jars, upgrade the docs build to Scala
2.12.8 and the main build to sbt 1.2.8.


dwijnand:
Reviewing I realise that I was confused because `mimaPreviousArtifacts` is empty on master.

But anyways let's make it less confusing by using `test.

PR #270, opened 2019-05-21
Upgrade to Scala 2.13.0-RC2

dwijnand:


play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -44,8 +44,8 @@ class JsonSpec extends org.specs2.mutable.Specification {
   implicit val FloatNumbersFormat: Format[FloatNumbers] = Json.format[FloatNumbers]
 
   implicit val PostFormat: Format[Post] = (
-    (__ \ 'body).format[String] and
-    (__ \ 'created_at).formatNullable[Option[Date]](
+    (__ \ Symbol("body")).format[String] and
```
marcospereira: Why not use the symbol literal anymore?

play-json/shared/src/test/scala/play/api/libs/json/JsonTransSpec.scala
```diff
@@ -163,14 +163,14 @@ class JsonTransSpec extends WordSpec with MustMatchers {
     "report the correct path in the JsError" when {
       "the field to modify doesn't exist" in {
         js.transform(
-          (__ \ 'field42).json.update(__.read[JsString])
-        ).asEither.left.get.head mustEqual ((__ \ 'field42, Seq(JsonValidationError("error.path.missing"))))
+          (__ \ Symbol("field42")).json.update(__.read[JsString])
+        ).asEither.swap.getOrElse(???).head mustEqual ((__ \ Symbol("field42"), Seq(JsonValidationError("error.path.missing"))))
```
marcospereira: Better to throw an error with a clear message instead of `???`. So, if the test fails, it will be easier to understand what is happening.

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
dwijnand: It's deprecated in Scala 2.13: https://github.com/scala/scala/pull/7395.

play-json/shared/src/test/scala/play/api/libs/json/JsonTransSpec.scala
dwijnand: Btw, it was throwing without a clear message before, with `.left.get`.

marcospereira:
By the way, this should be backported to 2.7.x, right?

dwijnand:
Yes.

PR #275, opened 2019-05-24
Target Java 8 in the build

marcospereira:
Also some scalac linting so that we can be aware of small code issues.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -65,6 +65,21 @@ scalaJSStage in ThisBuild := (sys.props.get("scalaJSStage") match {
   case _ => FastOptStage
 })
 
+val javacSettings = Seq(
+  "-source", "1.8",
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -65,6 +65,21 @@ scalaJSStage in ThisBuild := (sys.props.get("scalaJSStage") match {
   case _ => FastOptStage
 })
 
+val javacSettings = Seq(
+  "-source", "1.8",
+  "-target", "1.8",
```
dwijnand: This needs to be scoped in `javacOptions in (Compile, compile)` because otherwise `doc` (which creates javadoc) will pass it to `javadoc` which doesn't support it (sbt/sbt#1785).

dwijnand: https://github.com/playframework/play-json/pull/277

PR #292, opened 2019-06-28
Configure MiMa to check against 2.7.4 for Scala 2.13 builds

renatocaval:


build.sbt
```diff
@@ -46,14 +46,12 @@ def jsonDependencies(scalaVersion: String) = Seq(
...
 
-val previousVersions = Def.setting[Seq[String]] {
-  Seq("2.7.1")
-}
 
 def playJsonMimaSettings = mimaDefaultSettings ++ Seq(
   mimaPreviousArtifacts := {
-    if (scalaVersion.value.startsWith("2.13")) Set.empty
-    else previousVersions.value.map(organization.value %%% moduleName.value % _).toSet
+    val artifact = organization.value %%% name.value
```
renatocaval: I'm just copying this, but why are we using `%%%`? Isn't that for Scala.js?

cchantep: yep

PR #302, opened 2019-07-24
Added readDeepNullable and similar methods

### LINKED ISSUE #268: Optional reads if any part of the path is not present
Consider the following:

```
scala> (__ \ "a" \ "b").readNullable[String].reads(Json.obj("a" -> Json.obj()))
res1: play.api.libs.json.JsResult[Option[String]] = JsSuccess(None,)

scala> (__ \ "a" \ "b").readNullable[String].reads(Json.obj())
res2: play.api.libs.json.JsResult[Option[String]] = JsError(List((/a/b,List(JsonValidationError(List(error.path.missing),WrappedArray())))))
```

This is kind of surprising to me - since I would expect the above to behave the same as this, which it doesn't:

```
scala> (Json.obj() \ "a" \ "b").validateOpt[String]
res3: play.api.libs.json.JsResult[Option[String]] = JsSuccess(None,)
```

I'm not sure if there's a good reason why `readNullable` only returns a `JsSuccess(None)` if the last part of the path is missing, but fails for the rest. If there is a good reason, then it would be nice to have a version of it that would return `JsSuccess(None)` if any part of the path is missing. My use case for this is when you're only interested in a few small parts of a deeply nested object structure, so you want to extract out those few fields without having to incrementally `readNullable` each part of the path and then `flatMap` the reads (which ends up becoming very unreadable very quickly).

jroper:
Fixes #268

This variant of readNullable successfully reads deep paths even when nodes before the last are not present.

An alternative to this change might be to simply change the behavior of `readNullable`. It has been this way since https://github.com/playframework/playframework/pull/662, but there is no justification for why anywhere that I can find. Typically, `readNullable` is only used to read a path that is one level deep, so I don't imagine changing its behavior for deeper paths to be a big issue. I can't imagine any use case where you wouldn't `None` when nodes before the last path are not present.

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsConstraints.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsConstraints.scala
@@ -25,6 +25,11 @@ trait PathFormat {
   def nullableWithDefault[A](path: JsPath, defaultValue: => Option[A])(implicit f: Format[A]): OFormat[Option[A]] =
     OFormat(Reads.nullableWithDefault(path, defaultValue)(f), Writes.nullable(path)(f))
 
+  def deepNullable[A](path: JsPath)(implicit f: Format[A]): OFormat[Option[A]] =
+    OFormat(Reads.deepNullable(path)(f), Writes.nullable(path)(f))
+
+  def deepNullableWithDefault[A](path: JsPath, defaultValue: => Option[A])(implicit f: Format[A]): OFormat[Option[A]] =
+    OFormat(Reads.deepNullableWithDefault(path, defaultValue)(f), Writes.nullable(path)(f))
 }
 
 trait PathReads {
@@ -69,6 +74,42 @@ trait PathReads {
           case js => reads.reads(js).repath(path).map(Some(_))
         }
       ))
+    }
+
+  /**
+   * Reads a Option[T] search optional or nullable field at JsPath (field not found or null is None
+   * and other cases are Error).
+   *
+   * This method is designed for cherry-picking deeply nested fields whose path at any level may
+   * not exist.
+   *
+   * It runs through JsValue following all JsPath nodes on JsValue:
+   * - If any node in JsPath is not found => returns None
+   * - If any node in JsPath is found with value "null" => returns None
+   * - If the entire path is found => applies implicit Reads[T]
+   */
+  def deepNullable[A](path: JsPath)(implicit reads: Reads[A]): Reads[Option[A]] =
+    deepNullableWithDefault(path, None)
+
+  /**
+   * Reads an Option[T] search optional or nullable field at JsPath (field not found replaced by
+   * default value, null is None and other cases are Error).
+   *
+   * This method is designed for cherry-picking deeply nested fields whose path at any level may
+   * not exist.
+   *
+   * It runs through JsValue following all JsPath nodes on JsValue:
+   * - If any node in JsPath is not found => returns default value
+   * - If the last node in JsPath is found with value "null" => returns None
+   * - If the entire path is found => applies implicit Reads[T]
+   */
+  def deepNullableWithDefault[A](path: JsPath, defaultValue: => Option[A])(implicit reads: Reads[A]): Reads[Option[A]] =
+    Reads[Option[A]] { json =>
+      path.asSingleJson(json) match {
+        case JsDefined(JsNull) => JsSuccess(None)
+        case JsDefined(value) => reads.reads(value).repath(path).map(Some(_))
+        case JsUndefined() => JsSuccess(defaultValue)
+      }
     }
 
   def jsPick[A <: JsValue](path: JsPath)(implicit reads: Reads[A]): Reads[A] = at(path)(reads)
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsConstraints.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsConstraints.scala
@@ -24,12 +24,6 @@ trait PathFormat {
 
   def nullableWithDefault[A](path: JsPath, defaultValue: => Option[A])(implicit f: Format[A]): OFormat[Option[A]] =
     OFormat(Reads.nullableWithDefault(path, defaultValue)(f), Writes.nullable(path)(f))
-
-  def deepNullable[A](path: JsPath)(implicit f: Format[A]): OFormat[Option[A]] =
-    OFormat(Reads.deepNullable(path)(f), Writes.nullable(path)(f))
-
-  def deepNullableWithDefault[A](path: JsPath, defaultValue: => Option[A])(implicit f: Format[A]): OFormat[Option[A]] =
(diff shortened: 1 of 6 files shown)


lforite:
@mandubian any input on this ?

play-json/shared/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -71,6 +76,42 @@ trait PathReads {
       ))
     }
 
+  /**
+   * Reads a Option[T] search optional or nullable field at JsPath (field not found or null is None
```
renatocaval:
```suggestion
   * Reads an Option[T] search optional or nullable field at JsPath (field not found or null is None
```

play-json/shared/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -71,6 +76,42 @@ trait PathReads {
...
+   * It runs through JsValue following all JsPath nodes on JsValue:
+   * - If any node in JsPath is not found => returns None
+   * - If any node in JsPath is found with value "null" => returns None
+   * - If the entire path is found => applies implicit Reads[T]
+   */
+  def deepNullable[A](path: JsPath)(implicit reads: Reads[A]): Reads[Option[A]] =
+    deepNullableWithDefault(path, None)
+
+  /**
+   * Reads an Option[T] search optional or nullable field at JsPath (field not found replaced by
```
renatocaval:
```suggestion
   * Reads an Option[T] (optional or nullable field) at JsPath (field not found replaced by
```

play-json/shared/src/main/scala/play/api/libs/json/JsPath.scala
```diff
@@ -311,6 +311,34 @@ case class JsPath(path: List[PathNode] = List()) {
    */
   def readNullableWithDefault[T](defaultValue: => Option[T])(implicit r: Reads[T]): Reads[Option[T]] = Reads.nullableWithDefault[T](this, defaultValue)(r)
 
+  /**
+   * Reads a Option[T] search optional or nullable field at JsPath (field not found or null is None
```
renatocaval:
```suggestion
   * Reads an Option[T] (optional or nullable field) at JsPath (field not found or null is None
```

play-json/shared/src/main/scala/play/api/libs/json/JsPath.scala
```diff
@@ -439,6 +467,24 @@ case class JsPath(path: List[PathNode] = List()) {
     Format.nullableWithDefault[T](this, defaultValue)(f)
   }
 
+  /**
+   * Reads/Writes a Option[T] (optional or nullable field) at given JsPath
```
renatocaval:
```suggestion
   * Reads/Writes an Option[T] (optional or nullable field) at given JsPath
```

play-json/shared/src/main/scala/play/api/libs/json/JsPath.scala
```diff
@@ -311,6 +311,34 @@ case class JsPath(path: List[PathNode] = List()) {
...
+   *
+   * It runs through JsValue following all JsPath nodes on JsValue:
+   * - If any node in JsPath is not found => returns None
+   * - If any node in JsPath is found with value "null" => returns None
+   * - If the entire path is found => applies implicit Reads[T]
+   */
+  def readDeepNullable[T](implicit r: Reads[T]): Reads[Option[T]] = Reads.deepNullable[T](this)(r)
+
+  /**
+   * Reads an Option[T] search optional or nullable field at JsPath (field not found replaced by
```
renatocaval:
```suggestion
   * Reads an Option[T] (optional or nullable field) at JsPath (field not found replaced by
```

play-json/shared/src/main/scala/play/api/libs/json/JsPath.scala
```diff
@@ -439,6 +467,24 @@ case class JsPath(path: List[PathNode] = List()) {
...
+  /**
+   * Reads/Writes a Option[T] (optional or nullable field) at given JsPath
+   *
+   * @see JsPath.readDeepNullable to see behavior in reads
+   * @see JsPath.writeNullable to see behavior in writes
+   */
+  def formatDeepNullable[T](implicit f: Format[T]): OFormat[Option[T]] = Format.deepNullable[T](this)(f)
+
+  /**
+   * Reads/Writes a Option[T] (nullable field) at given JsPath
```
renatocaval:
```suggestion
   * Reads/Writes an Option[T] (nullable field) at given JsPath
```

play-json/shared/src/main/scala/play/api/libs/json/JsConstraints.scala
```diff
@@ -71,6 +76,42 @@ trait PathReads {
...
+   *
+   * This method is designed for cherry-picking deeply nested fields whose path at any level may
+   * not exist.
+   *
+   * It runs through JsValue following all JsPath nodes on JsValue:
+   * - If any node in JsPath is not found => returns default value
+   * - If the last node in JsPath is found with value "null" => returns None
+   * - If the entire path is found => applies implicit Reads[T]
+   */
+  def deepNullableWithDefault[A](path: JsPath, defaultValue: => Option[A])(implicit reads: Reads[A]): Reads[Option[A]] =
```
cchantep: Shouldn't it been the default behaviour for the existing nullable `Reads` ?

renatocaval: Yes, you are right. I'm ok to not introduce new methods and simply 'fix' the behaviour of the existing ones. 

@jroper, do you want to adapt your PR accordingly?

jroper:
@renatocaval I've updated to change the existing `readNullable` instead of introducing a new method.

Since I was changing such a central method, I wanted to ensure that I wasn't introducing a performance regression, so I created a new JMH benchmark that reads a case class with 3 optional fields from a JsValue that has two of them set. There was a significant performance increase, from 2650476 ops/s to 4155399 ops/s, a 56% performance improvement (note, this test is designed to execute that exact code and nothing else, so it's not going to have anywhere near as big an impact in the real world, but still its significant).

So realising that this code was really performance sensitive, and seeing that the code that I replaced it with, in the most common case, did a lot of allocations and unnecessary work, I decided to implement a fast path in  in `JsPath.asSingleJson` specifically targeted at `(__ \ "foo").readNullable[Foo]`, that is, reading from a path with a single key in it, which is by far the most common use of `readNullable`. Performance went up to 7569083 ops/second, a further 82% increase, for a total of 185% performance increase on the old implementation. Not bad!

Then I realised the same could be done for `(__ \ "foo").read[Foo]`, which uses `JsPath.asSingleJsonResult`. I didn't write a new test for this one, instead I reused the existing deserialise list one, which exercises a much broader set of code so we would expect the optimisation to have less of an impact (but be more meaningful), it went from 11664 ops/second to 16009 ops/second, a 37% performance increase.

PR #318, opened 2019-09-18
Handle null edgecases in JsValue

cchantep:


play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala
```diff
@@ -31,15 +31,24 @@ object JsValue {
...
  * Represents a Json boolean value.
  */
 sealed abstract class JsBoolean(
   val value: Boolean) extends JsValue with Product with Serializable {
 
-  def canEqual(that: Any): Boolean = that.isInstanceOf[JsBoolean]
+  def canEqual(that: Any): Boolean = that match {
+    case _: JsBoolean => true
+    case _ => false
+  }
```
jroper: Isn't `_.isInstanceOf[JsBoolean]` identical to `_: JsBoolean`? I would have thought they both compile to exactly the same byte code - `isInstanceOf` is a synthetic method that scalac will compile to a `instanceof` byte code instruction. Likewise, `_: JsBoolean` will also compile to an `instanceof` byte code instruction. Both are null safe. The new change is possibly slower because it will involve some jmp statements etc, rather than just directly returning the result of the `instanceof` instruction.

jroper:
I'm against this change.

For a start, all the `equals` methods now break the `hashCode`/`equals` contract, if two objects are equal, they must have the same hash code (otherwise you'll get very strange behavior in sets and maps), but now for example `JsNumber(null)` and `JsNull` are equal, and they don't have the same hash code.

Secondly, I don't think `JsNumber(null)` and `JsNull` should be considered equal. They are classes of different types. I don't think it even makes sense to have a `JsNumber(null)`, or any other class of nulls, null is not a bottom type in JSON, if something is of type `null` in JSON, it is not a number. If anything, `JsNumber(null)` should throw an NPE, it shouldn't be allowed. Explicitly allowing it also sets a dangerous precedent, if we support it in the code, then people might start using it, and then suddenly no one can rely on `.value` returning non null, it would mean if you want to use play-json safely, you have to null check every time you access a value.

cchantep:
@jroper Even if I agree `null` should never be used, there is now way to prevent `null` as inner value, which already lead to [misunderstanding](https://stackoverflow.com/questions/57958586/play-json-jsnull-is-not-equal-to-jsstringnull-how-to-get-around-this-issue/57959725).

jroper:
We can prevent `null` as an inner value:

```scala
case class JsString(value: String) {
  require(value != null, "JsString value must not be null")
}
```

That said, it's convention in Scala to never use nulls, and therefore assume everything is non null, unless providing a Java API or consuming a Java API, then you wrap nulls in Option as soon as practical and unwrap them as late as practical. Since play-json is a Scala only API, I don't see there being any need to handle nulls, the person posting to stack overflow about this needs to be educated on how to deal with optional values in Scala, because if they liberally use `orNull` on `Option`, play-json won't be the last place they'll have this problem, they'll run into problems all over the place as they pass null to libraries that assume null will never be passed.

cchantep:
For me, using `require` is more unsafe.

Even if that's not something recommended, there are still people ending with `null` there. And the cost to handle it as edgecase is not high there (issue w/ `equals`/`hashCode` can be fixed).

PR #319, opened 2019-09-19
Set size hint on builder in traversable reads

jroper:
Very minor optimisation that I noticed could be done when looking over the code for some other reason.


jroper:
Note, this is even less useful than it sounds - none of Scalas immutable collections implement sizeHint (they don't need to, they allocate as needed with no buffer copy penalty), the only collection that does implement it is `ArrayBuffer`.

PR #333, opened 2019-09-30
Update Travis Jabba settings

renatocaval:
We have a few builds failing on Travis because of adopt@1.8.0-202-08. 

Strange enough, this was working before. This PR brings a more recent version of it and hopefully, Travis will stop to complain.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,12 @@
 language: scala
 
-before_install:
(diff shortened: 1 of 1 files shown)


mkurz:
@renatocaval See https://github.com/playframework/playframework/pull/9688

renatocaval:
Thanks @mkurz, I missed those fixes. I will apply it here as well.

PR #340, opened 2019-10-25
[2.7.x]: Use latest adoptOpenJDK

ignasi35:


ignasi35:
see https://github.com/lightbend/mima/issues/422

PR #344, opened 2019-11-08
Validate documentation code and Scaladoc links

marcospereira:
Reference documentation was not validated for play-json.

This does some build reorganization and also add new commands in Travis to do validation for the documentation code and links.


build.sbt
```diff
@@ -25,7 +25,10 @@ val specsBuild = Def.setting[Seq[ModuleID]] {
     case _ => "4.5.1"
   }
 
-  Seq("org.specs2" %% "specs2-core" % specsVersion)
+  Seq(
+    "org.specs2" %% "specs2-core" % specsVersion,
+    "org.specs2" %% "specs2-junit" % specsVersion,
```
marcospereira: To fix warnings about JUnit report generation.

build.sbt
```diff
@@ -261,6 +264,29 @@ lazy val benchmarks = project
   .settings(commonSettings)
   .dependsOn(`play-jsonJVM`)
 
+lazy val docs = project
```
marcospereira: A sub-project instead of a separated project with its `build.sbt`, `project/`, etc.

project/plugins.sbt
```diff
@@ -7,11 +7,14 @@ resolvers += Resolver.typesafeRepo("releases")
 
 addSbtPlugin("com.typesafe.play" % "interplay" % sys.props.get("interplay.version").getOrElse("2.1.1"))
 
+addSbtPlugin("com.typesafe.play" % "play-docs-sbt-plugin" % sys.props.getOrElse("play.version", "2.8.0-RC1"))
```
marcospereira: Unfortunately, this creates a circular dependency with Play, something that already exists for Twirl:

https://github.com/playframework/twirl/blob/b712905b2ff8877b9fcf5fec8ec8f247a1ec5ea6/docs/project/plugins.sbt#L7

A tolerable problem though, since this does not needs to be up to date with Play releases.

PR #352, opened 2019-11-19
[2.8.x]: Fix release process

marcospereira:
Two major problems:

1. It was not promoting the staged repository
2. It was generating multiple staging repositories

This should fix both by using the defaults provided by interplay.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,21 +3,21 @@
  */
 import sbt._
 import sbt.util._
-import scala.sys.process._
 import sbt.io.Path._
 
 import interplay.ScalaVersions
-import ReleaseTransformations._
 
-import com.typesafe.tools.mima.core._
 import com.typesafe.tools.mima.plugin.MimaPlugin.mimaDefaultSettings
 import com.typesafe.tools.mima.plugin.MimaKeys.mimaBinaryIssueFilters
 import com.typesafe.tools.mima.plugin.MimaKeys.mimaPreviousArtifacts
 
-import sbtcrossproject.crossProject
+import sbtcrossproject.CrossPlugin.autoImport.crossProject
 import sbtcrossproject.CrossType
 
 resolvers ++= DefaultOptions.resolvers(snapshot = true)
+
+playBuildRepoName in ThisBuild := "play-json"
+publishTo in ThisBuild := sonatypePublishToBundle.value
 
 val specsBuild = Def.setting[Seq[ModuleID]] {
   val specsVersion = CrossVersion.partialVersion(scalaVersion.value) match {
@@ -104,12 +104,6 @@ lazy val commonSettings = Def.settings(
     // Filtering tests that are not stable in Scala 2.13 yet.
     Tests.Argument(TestFrameworks.ScalaTest, "-l", "play.api.libs.json.UnstableInScala213")
   ),
-  publishTo := Some(
-    if (isSnapshot.value)
-      Opts.resolver.sonatypeSnapshots
-    else
-      Opts.resolver.sonatypeStaging
-  ),
   headerLicense := {
     val currentYear = java.time.Year.now(java.time.Clock.systemUTC).getValue
     Some(
@@ -132,7 +126,7 @@ lazy val commonSettings = Def.settings(
 
 lazy val root = project
   .in(file("."))
-  .enablePlugins(PlayRootProject, ScalaJSPlugin)
+  .enablePlugins(PlayRootProject, ScalaJSPlugin, PlayNoPublish)
   .aggregate(
     `play-jsonJS`,
     `play-jsonJVM`,
@@ -140,10 +134,7 @@ lazy val root = project
     `play-functionalJVM`,
     `play-json-joda`
   )
-  .settings(
-    commonSettings,
-    publishTo := None
-  )
+  .settings(commonSettings)
 
 lazy val `play-json` = crossProject(JVMPlatform, JSPlatform)
   .crossType(CrossType.Full)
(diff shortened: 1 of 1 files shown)


renatocaval:
@marcospereira, something went wrong in Travis. Build is failing with error in build definition.

ignasi35:
> @marcospereira, something went wrong in Travis. Build is failing with error in build definition.

```
Failed to sort List(
  interplay.PlayNoPublishBase, 
  org.scalajs.sbtplugin.ScalaJSPlugin, 
  interplay.PlayRootProjectBase, 
  sbt.plugins.CorePlugin, 
  interplay.PlaySonatypeBase, 
  sbt.plugins.JvmPlugin, 
  interplay.PlayBuildBase, 
  interplay.PlayReleaseBase, 
  sbt.plugins.IvyPlugin, 
  sbt.plugins.Giter8TemplatePlugin, 
  scalajscrossproject.ScalaJSCrossPlugin, 
  interplay.PlayBintrayBase, 
  de.heikoseeberger.sbtheader.HeaderPlugin, 
  org.scalafmt.sbt.ScalafmtPlugin, 
  org.scalajs.sbtplugin.internal.ScalaJSGlobalPlugin, 
  sbtwhitesource.WhiteSourcePlugin, 
  com.typesafe.tools.mima.plugin.MimaPlugin, 
  org.portablescala.sbtplatformdeps.PlatformDepsPlugin, 
  sbt.plugins.SemanticdbPlugin, 
  sbt.plugins.JUnitXmlReportPlugin, 
  sbtcrossproject.CrossPlugin, 
  com.jsuereth.sbtpgp.SbtPgp, 
  sbtrelease.ReleasePlugin, 
  interplay.PlayWhitesourcePlugin) 
topologically
```
Sounds like a [cyclical dependency on the plugins](https://travis-ci.com/playframework/play-json/jobs/258409826#L270).

build.sbt
```diff
@@ -132,18 +126,15 @@ lazy val commonSettings = Def.settings(
 
 lazy val root = project
   .in(file("."))
-  .enablePlugins(PlayRootProject, ScalaJSPlugin)
+  .enablePlugins(PlayRootProject, ScalaJSPlugin, PlayNoPublish)
```
ignasi35: (I'll ignore `ScalaJSPlugin`)

The `enablePlugins` above can be expanded to:

1. `PlayRootProject` expands to `PlayRootProjectBase &&! PlaySbtCompat.optScriptedAutoPlugin`
1. `PlayNoPublish` expands to `PlayNoPublishBase &&! BintrayPlugin &&! Sonatype`

Then we have:

```
enablePlugins(
  PlayRootProjectBase &&! PlaySbtCompat.optScriptedAutoPlugin, // exclude(ScriptedAutoPlugin)
  PlayNoPublishBase &&! BintrayPlugin &&! Sonatype // exclude(BintrayPlugin, Sonatype)
)
```

Now, `PlayRootProjectBase` brings some stuff transitively:

```
object PlayRootProjectBase extends AutoPlugin {
  override def trigger = noTrigger
  override def requires = PlayBuildBase && PlayBintrayBase && PlaySonatypeBase && PlayReleaseBase
```

So the end result looks like:

```
enablePlugins(
  PlayRootProjectBase &&! PlaySbtCompat.optScriptedAutoPlugin, // exclude(ScriptedAutoPlugin)
  PlayBuildBase, 
  PlayBintrayBase,  // brings `BintrayPlugin` transitively
  PlaySonatypeBase, // brings `Sonatype` transitively
  PlayReleaseBase,

  PlayNoPublishBase &&! BintrayPlugin &&! Sonatype // exclude(BintrayPlugin, Sonatype)
  PlayBuildBase,    // transitively added by `PlayNoPublishBase`
)
```

Which, I guess, looks like:

```
enablePlugins(
  PlayRootProjectBase
  PlayBuildBase, 
  PlayReleaseBase,
  PlayNoPublishBase
)
```

- - - - - - - - - - - 

IIUC, the purpose is to not publish the `root` project. In that case, we don't want `PlayReleaseBase`. 

Then, we could be tempted to also remove `PlayRootProjectBase` (which is the one bringing in `PlayReleaseBase` but `PlayRootProjectBase` is defining the [`crossScalaVersions`](https://github.com/playframework/interplay/blob/master/src/main/scala/interplay/PlayBuildBase.scala#L356-L362))

I don't know what I'm talking about anymore.

 - - - - - - - - - - 

OMG, this is so confusing.

According to [the docs](https://github.com/playframework/interplay#the-root-project) this aggregating `root` project should be a `PlayRootProject` and then, each of the aggregated projects that must not be published should be `PlayNoPublish`.

Ofc, `interplay` being created before sbt `1.x` it's possible that some assumptions no longer hold true in the sbt `1.3+` ecosystem.

ignasi35: I think the `NoPublish` is unnecessary and your change above:

```

  publishTo := Some(
    if (isSnapshot.value)
      Opts.resolver.sonatypeSnapshots
    else
      Opts.resolver.sonatypeStaging
  ),
```

is enough.

build.sbt
```diff
@@ -126,7 +126,7 @@ lazy val commonSettings = Def.settings(
 
 lazy val root = project
   .in(file("."))
-  .enablePlugins(PlayRootProject, ScalaJSPlugin, PlayNoPublish)
```
marcospereira: As `PlayNoPublish` was removed, we need to add the settings manually. I think `publish / skip := true` should be enough (as [documented here](https://www.scala-sbt.org/1.x/docs/Publishing.html#Skip+publishing)) but maybe re-add `publishTo: None` as well.

ignasi35: `publish / skip := true` is what we use in other projects (eg https://github.com/playframework/play-grpc/blob/master/project/Publish.scala#L13).

I've ran `show publishTo` with this setup and the `root` project reports no target (which is what I think we wanted).

How about we merge and we revisit this on the next release?

PR #172, opened 2018-08-08
Fix javadoc for lazyWriteNullable and lazyWrite

kapkaev:
## Fixes

Fix javadoc for [lazyWriteNullable](https://github.com/playframework/play-json/blob/master/play-json/shared/src/main/scala/play/api/libs/json/JsPath.scala#L391-L395) and [lazyWrite](https://github.com/playframework/play-json/blob/master/play-json/shared/src/main/scala/play/api/libs/json/JsPath.scala#L373-L377)

PR #363, opened 2019-12-02
Fix code examples in Scaladoc

cchantep:
Make sure all the code examples in the Scaladoc `{{{ ... }}}` are consistent (compilable).


cchantep:
Also fix #172

PR #369, opened 2019-12-07
Add iterableWrites2 because iterableWrites is incompatible with Map[A, B]

### LINKED ISSUE #368: 2.8.0 regression: iterableWrites is not applicable to Map[A, B]
### Play JSON Version
2.8.0

### API
Scala

### Operating System
MacOS 10.14.6

### JDK
```
openjdk version "11.0.2" 2019-01-15
OpenJDK Runtime Environment 18.9 (build 11.0.2+9)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.2+9, mixed mode)
```

### Expected Behavior

Given a map with non-String keys I expect it to be written as an array of two-element arrays – that's how it works in play-json 2.7.4.

E.g. this test passes
```scala
Json.toJson(Map(1 -> "one")).mustEqual(Json.arr(Json.arr(1, "one")))
```

### Actual Behavior

Instead, the code fails to compile
```
No Json serializer found for type scala.collection.immutable.Map[Int,String]. Try to implement an implicit Writes or Format for this type.
```

New imiplicit `iterableWrites`
```scala
implicit def iterableWrites[A, M[T] <: Iterable[T]](implicit w: Writes[A]): Writes[M[A]]
```
can't keep the exact type of `Map[A, B]` since the shape is different (`_[_, _]` vs `_[_]`) and `Writes[A]` is not covariant anymore (it was in 2.7.4).

### Possible fix

Unless I'm missing something, a bit different signature should do the trick because it keeps the exact type of `I`.
```
implicit def iterableWrites2[A, I](implicit ev: I <:< Iterable[A], w: Writes[A]): Writes[I]
```

I will post a PR soon.

sirmax:
# Pull Request Checklist

## Fixes

Fixes #368 

## Purpose

Add new implicit `iterableWrites2` with a signature that can work with arbitrary shapes.
Make old `iterableWrites` non-implicit to let new implicit work, but keep it as deprecated to preserve backward compatibility.

## Background Context

Took this approach because it worked for me :)

## References

None.

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
@@ -428,11 +428,20 @@ sealed trait LowPriorityWrites extends EnvWrites {
 
   /**
    * Serializer for Iterable types.
+   *
+   * Deprecated due to incompatibility with non `_[_]` shapes, #368.
    */
-  implicit def iterableWrites[A, M[T] <: Iterable[T]](implicit w: Writes[A]): Writes[M[A]] = {
+  @deprecated("Use `iterableWrites2`", "2.8.1")
+  def iterableWrites[A, M[T] <: Iterable[T]](implicit w: Writes[A]): Writes[M[A]] =
+    iterableWrites2[A, M[A]]
+
+  /**
+   * Serializer for Iterable types.
+   */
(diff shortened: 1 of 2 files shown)


renatocaval:
@dwijnand / @marcospereira, could you have a look at this?

sirmax:
> I'm probably being over-cautious.

It's OK. Better safe than sorry :)

I think I should add details to `@deprecated` annotation and re-use new method from the deprecated one. I'll update the PR in a moment.

PR #377, opened 2019-12-16
Fix supported scala versions in README

renatocaval:


README.md
```diff
@@ -26,7 +26,7 @@ To get started, you can add play-json as a dependency in your project:
   </dependency>
   ```
  
-Play JSON supports Scala 2.10, 2.11 and 2.12. Choosing the right JAR is automatically managed in sbt. If you're using Gradle or Maven then you need to use the correct version in the `artifactId`.
+Play JSON supports Scala 2.11, 2.12 and 2.13. Choosing the right JAR is automatically managed in sbt. If you're using Gradle or Maven then you need to use the correct version in the `artifactId`.
```
renatocaval: The readme is mentioning 2.10, 2.11 and 2.12, but the 2.7.x build cross builds to 2.11, 2.12 and 2.13.

PR #380, opened 2019-12-16
Enable sbt-dynver

renatocaval:
Targeting 2.8.x with the intention to merge back in `master`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -86,6 +86,39 @@ libraryDependencies in ThisBuild ++= Seq(
   ("com.github.ghik" % "silencer-lib" % silencerVersion % Provided).cross(CrossVersion.full)
 )
 
+// Customise sbt-dynver's behaviour to make it work with tags which aren't v-prefixed
+dynverVTagPrefix in ThisBuild := false
+
+// Sanity-check: assert that version comes from a tag (e.g. not a too-shallow clone)
+// https://github.com/dwijnand/sbt-dynver/#sanity-checking-the-version
+Global / onLoad := (Global / onLoad).value.andThen { s =>
+  val v = version.value
+  if (dynverGitDescribeOutput.value.hasNoTags)
+    throw new MessageOnlyException(
+      s"Failed to derive version from git tags. Maybe run `git fetch --unshallow`? Version: $v"
+    )
+  s
+}
+
+// this is overrides interplay release settings with some adaptations for working with sbt-dynver
+// moreover, it contain only bits necessary for play-json, nothing else
+lazy val releaseSettings: Seq[Setting[_]] = Seq(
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -86,6 +86,39 @@ libraryDependencies in ThisBuild ++= Seq(
...
+    throw new MessageOnlyException(
+      s"Failed to derive version from git tags. Maybe run `git fetch --unshallow`? Version: $v"
+    )
+  s
+}
+
+// this is overrides interplay release settings with some adaptations for working with sbt-dynver
+// moreover, it contain only bits necessary for play-json, nothing else
+lazy val releaseSettings: Seq[Setting[_]] = Seq(
+  releasePublishArtifactsAction := PgpKeys.publishSigned.value,
```
marcospereira: I thought that it was only necessary to override `releaseProcess`. Are the others really necessary here?

build.sbt
```diff
@@ -86,6 +86,39 @@ libraryDependencies in ThisBuild ++= Seq(
...
+Global / onLoad := (Global / onLoad).value.andThen { s =>
+  val v = version.value
+  if (dynverGitDescribeOutput.value.hasNoTags)
+    throw new MessageOnlyException(
+      s"Failed to derive version from git tags. Maybe run `git fetch --unshallow`? Version: $v"
+    )
+  s
+}
+
+// this is overrides interplay release settings with some adaptations for working with sbt-dynver
```
marcospereira:
```suggestion
// this overrides interplay release settings with some adaptations for working with sbt-dynver
```

build.sbt
renatocaval: It's true. We only need to overwrite `releaseProcess`. 

The original [code](https://github.com/playframework/interplay/blob/master/src/main/scala/interplay/PlayReleaseBase.scala#L40-L61) in interplay had some embedded functions that were used to define the process. 

I started by copying everything and then while cleaning up, I remove all of them. It's not needed any more, we can fall back to only re-defining the `releaseProcess`.

renatocaval: @marcospereira, this is updated now

build.sbt
```diff
@@ -127,6 +141,22 @@ lazy val root = project
...
   .settings(commonSettings)
+  .settings(
+    Seq(
+      // this overrides releaseProcess to make it work with sbt-dynver
+      releaseProcess := {
+        import ReleaseTransformations._
+        Seq[ReleaseStep](
+          checkSnapshotDependencies,
+          runClean,
+          releaseStepCommandAndRemaining("+test"),
```
ignasi35:
```suggestion
          releaseStepCommandAndRemaining("+test"), // <- this needs to be removed when releasing from Travis
```

😂

PR #432, opened 2020-03-05
Removed Joda functions and dependency from base play-json library.

bursauxa:
# Purpose

```scala
// TODO: remove joda after 2.6.0
```

I think the time has come :slightly_smiling_face:

# Background Context

play-json-joda has been there for quite some time now, as well as Java 8 Time.

Meanwhile, many projects without use of Joda are still unknowingly downloading Joda through Play-JSON or even more indirectly through Play.

If people maintaining this library are in agreement, I believe we can today move forward with cutting the Joda dependency from the base library, so it stays strictly in play-json-joda.


bursauxa:
Note that the build is failing but now that the SBT format is good, I think what's left is a "normal error":

```Travis
java.lang.RuntimeException: Failed binary compatibility check against com.typesafe.play:play-json_2.12:2.8.1! Found 48 potential problems
```

It's reporting that I removed methods, which is indeed the intent.

gmethvin:
You're right, this probably should have been removed in 2.7 or 2.8. We can merge this, but since we're removing methods we might not want to backport it to 2.8. Or maybe since these are deprecated methods we can make an exception. I'm not sure.

ignasi35:
> Note that the build is failing but now that the SBT format is good, I think what's left is a "normal error":
> 
> ```
> java.lang.RuntimeException: Failed binary compatibility check against com.typesafe.play:play-json_2.12:2.8.1! Found 48 potential problems
> ```
> 
> It's reporting that I removed methods, which is indeed the intent.

Yeah, you will have to setup the MiMa exclusions. See the [recommended](https://travis-ci.com/github/playframework/play-json/jobs/295053665#L1847) settings that need to be included and test that locally using `sbt mimaReportBinaryIssues`.

ignasi35:
> You're right, this probably should have been removed in 2.7 or 2.8. We can merge this, but since we're removing methods we might not want to backport it to 2.8. Or maybe since these are deprecated methods we can make an exception. I'm not sure.

I'd rather not backport.

bursauxa:
It's probably safer to keep it for 2.9 indeed, but I am not sure what the schedule for that is... If it is far in the distance, or worse if 2.8 is the last major version for Scala 2, then maybe make an exception.

ennru:
Thank you @bursauxa for starting this PR!

PR #434, opened 2020-03-25
Introduce method `of` and `pure` specialized for `OWrites`

avdv:
# Pull Request Checklist

## Purpose

This change allows us to resolve an `OWrites[T]` via `OWrites.of[T]`.

## Background Context

I was trying to use `contramap` on an `OWrites[T]`:
```scala
OWrites.of[T].contramap[X](_.whatever)
```
Surpisingly, this led to an error later, since the caller expected an `OWrites[X]` but `OWrites.of` is simply inherited from `ConstraintWrites` and returns a `Writes`.

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
@@ -77,6 +77,13 @@ trait OWrites[A] extends Writes[A] {
 object OWrites extends PathWrites with ConstraintWrites {
   import play.api.libs.functional._
 
+  def of[A](implicit w: OWrites[A]): OWrites[A] = w
+
+  def pure[A](fixed: => A)(implicit wrs: OWrites[A]): OWrites[JsValue] =
+    OWrites[JsValue] { js =>
+      wrs.writes(fixed)
+    }
+
   /**
    * An `OWrites` merging the results of two separate `OWrites`.
    */


cchantep:
That change would break the compat.

avdv:
> That change would break the compat.

Yep, I also noticed this when the build failed. Basically this means that this could only be done in major version change, right?

raboof:
> Basically this means that this could only be done in major version change, right?

There might be a way to achieve what this PR does while maintaining binary compatibility - for example by keeping the original method, but making it `private[json]`. That keeps the method in the compiled classes, but shields it from new users.

I'm not sure I understand what this change is meant to achieve, though - perhaps it would be good to add a testcase demonstrating this?

cchantep:
> Yep, I also noticed this when the build failed. Basically this means that this could only be done in major version change, right?

Theorically, yes

avdv:
> There might be a way to achieve what this PR does while maintaining binary compatibility - for example by keeping the original method, but making it `private[json]`. That keeps the method in the compiled classes, but shields it from new users.
@raboof Thanks for you input on this, I went down that direction quite a bit, only to realize that it is a lot easier to maintain binary compat without any trickery. :smile: 

> I'm not sure I understand what this change is meant to achieve, though - perhaps it would be good to add a testcase demonstrating this?

Currently, all the inherited methods from `ConstraintWrites` in `OWrites` are superfluous since you could simply call the same methods from `Writes` or even `Writes.constraints`.

Also, some of the methods are really counter-intuitive  or "misleading" being defined on `OWrites`:

* `OWrites.optionWithNull` -- cannot return an OWrites since it might write `JsNull`
* `OWrites.pruned` -- always writes `JsNull`, so cannot ever return an `OWrites`

Why are these part of the `OWrites` object at all? They are not specific to `OWrites`' usage nor useful in this context.

Otherwise, there are methods `of` and `pure` that just return a `Writes`, even if -- in the context of `OWrites` -- should better return an `OWrites`.

So, instead of writing `implicitly[OWrites[X]]` one can use `OWrites.of[X]` in the same way someone would expect for `Writes` and `Reads`.

PR #454, opened 2020-05-06
Update to scalajs 1.0.1

mliarakos:
# Pull Request Checklist

* [x] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?

## Fixes

Fixes #433.

## Purpose

This PR updates to Scala.js 1.0.1 and provides a fix for the missing MiMA previous artifacts.

## Background Context

I made the MiMA previous artifacts setting conditioned on the environment and the previous `play-json` version. It should only apply to the JS platform and for the next release.

## References

This PR was motivated by the discussion around #433.

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -15,7 +15,7 @@ addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.7.0")
 
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.6.0")
 
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.1")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.1.0")
 
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.0.0")
 
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -15,7 +15,7 @@ addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.7.0")
 
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.6.0")
 
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.1.0")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.1")
 
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.0.0")
 
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -15,7 +15,7 @@ addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.7.0")
 
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.6.0")
 
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.33")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.1")
 
 addSbtPlugin("org.portable-scala" % "sbt-scalajs-crossproject" % "1.0.0")
 
--- a/build.sbt
+++ b/build.sbt
@@ -44,8 +44,8 @@ def jsonDependencies(scalaVersion: String) = Seq(
 def playJsonMimaSettings = Seq(
   mimaPreviousArtifacts := ((crossProjectPlatform.?.value, previousStableVersion.value) match {
     case (Some(JSPlatform), Some("2.8.1")) => Set.empty
-    case (_, Some(previousVersion))             => Set(organization.value %%% moduleName.value % previousVersion)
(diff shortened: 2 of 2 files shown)


thebridsk:
Shouldn't the test be is scalajs and 2.8.1 and scalajsversion starts with 1.0

mliarakos:
I don't think we need to check the Scala.js version, but I might be wrong. Here's my reasoning: This PR removes support for Scala.js 0.6 and moves to Scala.js 1.0 only (i.e., we're not publishing artifacts for both versions). This means the Scala.js version can only be 1.0 and we don't have to check for 0.6.

I also assumed the MiMA check would be removed after the next release because we will have previous artifacts once it's released. We don't need to future proof the check because it's going away. Basically, this PR is a bridge to get `play-json` to Scala.js 1.0 and it can go back to the previous logic for `mimaPreviousArtifacts` after making the transition.

Feedback welcome, especially if I have faulty assumptions.

ennru:
I'd suggest we release 2.8.2 for Scala.js 0.6 and do 2.9.0 for 1.0 just after it.

ennru:
Ah, this is co-versioned with Play. Need to discuss with people who know how we best do it...

an-tex:
> Ah, this is co-versioned with Play. Need to discuss with people who know how we best do it...

@ennru is it really? I somehow had it mind it was made independent and looking at the releases: play-json:2.8.1 was released on 12 Dec 2019 vs play:2.8.1 on 1 Feb 2020.

ennru:
> @ennru is it really?

No, I'm not sure. I'm not very familiar with the myriad of Play repos.

And the dimension of Scala.js things, doesn't make it easier. Would this need to cross-build for Scala.js 0.6, 1.0 and 1.1?

sjrd:
> And the dimension of Scala.js things, doesn't make it easier. Would this need to cross-build for Scala.js 0.6, 1.0 and 1.1?

Only 0.6.x and 1.x (whichever). 1.0 and 1.1 are the same thing. See https://users.scala-lang.org/t/announcing-scala-js-1-1-0/6053/3?u=sjrd

ennru:
Thank you @sjrd for jumping in. After I commented I had the chance to read the Scala.js release notes...

So we'd want to cross publish this to Scala.js 0.6 and 1.0.

an-tex:
I guess what @mliarakos was suggesting is to drop sjs0.6 support for the next release. Many libraries started going down this route due to the additional effort and the [high rate of adoption of sjs1](https://github.com/scala-js/scala-js/issues/4045). I tend to agree

ennru:
I'm all for dropping Scala.js 0.6, if the Scala.js community is ready for it...

My ambition is to make publishing this library easier so that it becomes easier to get out even the minor changes that it sees nowadays.

mliarakos:
I rebased and bumped Scala.js to 1.1.0.

I can re-work the PR to cross publish for Scala.js 0.6 and 1.x, but it will complicate the build process. The [recommended approach](https://www.scala-js.org/news/2020/02/25/announcing-scalajs-1.0.0/) to cross build uses an environment variable. I'll also have to update the Travis config and we'll need to make sure the release/publish process uses the environment variable (I'm not sure how releases are handled for `play-json`).

As @an-tex mentioned, the community seems to be moving quickly towards Scala.js 1.x. However, if we want a more measured transition we could cross publish for `play-json` 2.8.x and then drop Scala.js 0.6 support in `play-json` 2.9. I prefer the immediate cut over, but I leave that up to the maintainers.

raboof:
I would be fine with merging this without cross-building and then releasing it as 2.9.0

project/plugins.sbt
```diff
@@ -19,7 +19,7 @@ addSbtPlugin("com.jsuereth" % "sbt-pgp" % "2.0.1")
 
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.6.0")
 
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.33")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.1.0")
```
ennru, line 22: > "It is not forward binary compatible with 1.0.x: libraries compiled with 1.1.0 cannot be used with 1.0.x."

https://users.scala-lang.org/t/announcing-scala-js-1-1-0/6053/3?u=sjrd

So it would be better to stay on 1.0.1 here and be compatible with both Scala.js 1.0 and 1.1.

ennru:
I rebased this and kept Scala.js 1.0.1
My plan would be to release Play JSON 2.9.0 with this merged very soon.

mliarakos:
Thanks for handling the rollback to Scala.js 1.0.1 @ennru.

It'll be handled in a follow-on PR, but is there a preference on when we update to Scala.js 1.1.x?

ennru:
> but is there a preference on when we update to Scala.js 1.1.x?

What's the benefit of going to Scala.js 1.1.x and dropping 1.0.x?

Open an issue to discuss the pros and cons, please.

PR #464, opened 2020-05-20
Merge 2.8.x into master

ennru:
The difference is minimal and we could start releasing from master instead.


ennru:
I'll do it locally instead.

PR #469, opened 2020-05-25
Travis: use stages

ennru:
Introduce stages to test first, publish after it


.travis.yml
```diff
@@ -18,6 +21,21 @@ script:
   - sbt -DscalaJSStage=full ++$TRAVIS_SCALA_VERSION validateCode test mimaReportBinaryIssues publishLocal
   - sbt ++$TRAVIS_SCALA_VERSION docs/test docs/validateDocs
 
+jobs:
+  include:
+    - stage: publish
+      name: "TEMPORARY, will become Publish artifacts to Bintray"
+      script: sbt +mimaReportBinaryIssues
+      scala: 2.13.1
```
renatocaval, line 29: why do we need the scala version?

ennru: Otherwise the Travis Matrix kicks in and runs it for 2.12 and 2.13.

renatocaval: I think you lost me? Are you saying that the `scala` section on line 13 will cause it to run twice?

I think we should remove this. The build.sbt already define the default scala version and the the `+` will take care of the cross building. 

Btw, play-soap doesn't have it: https://github.com/playframework/play-soap/blob/master/.travis.yml

renatocaval: We don't have a definition for state `test` in this file? Is that a default staget? And will it pick what is defined in `script` (line 20)

ennru: Play SOAP doesn't use a Travis Matrix.

Yes, `test` automatically inherits the global `script` as it doesn't have its own.

renatocaval: That was my point. I was suggesting to remove the mix of matrix with stages.  Let's only have stages, otherwise we need extra brain cycles to map the effect they have on each other.

ennru: I unrolled it now.

dwijnand:
What's the motivation behind the matrix unroll?  Easier to grok?  Easier to vary things for different versions (at the expense of consistency)?

ennru:
Easier to grok, fewer WTFs, see what is there.

dwijnand:
Gotcha.

Btw, I just remembered, there's also https://config.travis-ci.com/explore to help understand the matrix expansion as well as other details.

PR #495, opened 2020-08-07
Close #479 - scalacheck-1-14 3.2.0.0 (was 3.1.2.0) 

cchantep:


ennru:
See #498, we need to stay on Scalatest 3.1 for now AFAICS.

cchantep:
Updated

PR #512, opened 2020-09-11
Support multi-level sealed trait in macros

cchantep:
Considering:

```scala
  sealed trait TreeValue

  sealed trait SubLevel extends TreeValue

  case class Leaf1(value: String) extends TreeValue
  case class Leaf2(value: Int) extends SubLevel

  object TreeValue {
    private implicit val leaf1: OFormat[Leaf1] = Json.format
    private implicit val leaf2: OFormat[Leaf2] = Json.format
    private implicit val subLevel: OFormat[SubLevel] = Json.format

    implicit val format: OFormat[TreeValue] = Json.format
  }
```

For now the discriminator from the top level overwrite the one from the sub one:

```scala
// So:
Json.toJson[TreeValue](Leaf2(1)) == Json.obj("_type" -> "SubLevel", "value" -> 1)

// Instead:
Json.toJson[TreeValue](Leaf2(2)) == Json.obj("_type" -> "Leaf2", "value" -> 2)
```

*See [failing test](https://travis-ci.com/github/playframework/play-json/jobs/383684122#L1087)*

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
@@ -676,7 +676,7 @@ class JsMacroImpl(val c: blackbox.Context) {
               case jsv => $json.JsObject(Seq("_value" -> jsv))
             }
 
-            jso + ($config.discriminator -> $json.JsString($config.typeNaming(${t.typeSymbol.fullName})))
+            $json.JsObject(Map($config.discriminator -> $json.JsString($config.typeNaming(${t.typeSymbol.fullName})))) ++ jso
           }"""
           }
         )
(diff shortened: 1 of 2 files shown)


play-json/shared/src/main/scala/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -676,7 +676,7 @@ class JsMacroImpl(val c: blackbox.Context) {
               case jsv => $json.JsObject(Seq("_value" -> jsv))
             }
 
-            jso + ($config.discriminator -> $json.JsString($config.typeNaming(${t.typeSymbol.fullName})))
+            $json.JsObject(Map($config.discriminator -> $json.JsString($config.typeNaming(${t.typeSymbol.fullName})))) ++ jso
```
cchantep, line 679: If the sub-writer use the same discriminator name, let it overwrite the value with the more specific type

ignasi35 (commented):
I like the improvement but I think it has the potential of breaking some existing code that relies on this behavior. 
Should this be released on a new major or maybe behind a setting?

cchantep:
> I like the improvement but I think it has the potential of breaking some existing code that relies on this behavior.
> Should this be released on a new major or maybe behind a setting?

For me the fix make sure the already expected behaviour, so that's not breaking.
Anyway if we think about a setting, it would mean exposing a setting to keep an invalid discriminator ...

cchantep:
ping

ignasi35:
Yeah, I think you're right. The improvement on the PR makes everything work as one would expect. If someone relies on the current (unexpected) behavior, then they can force their way out of the correct implementation by tunning the Format's.

ignasi35 (approved):

PR #533, opened 2020-11-02
Add functional factories for KeyReads and KeyWrites typeclasses

cchantep:


ignasi35 (approved):
Oh! This is an intersting improvement! 👍

PR #538, opened 2020-11-08
Add documentation and examples about Reads/Writes/Format combinators

cchantep:


ignasi35 (approved):
I love it! Mentioning this feature in the docs and adding a few snippets demo-ing it will make it a lot more discoverable than relying on IDE's autocompletion.

PR #573, opened 2021-02-18
dotty remove unnecessary constraints

jtjeferreira:


dwijnand:
Initial thread: https://github.com/playframework/play-json/pull/572#discussion_r578435998

I wonder if this works because it just keeps summoning new instances instead of reusing?

jtjeferreira:
> I wonder if this works because it just keeps summoning new instances instead of reusing?

I see your point. Any idea how I can write an unit test for that?

dwijnand:
Hmm, perhaps just do something like this?

```scala
object Foo {
  private val counter = AtomicInteger(0)
  implicit def reads: Reads[Foo] = {
    counter.inc()
    Json.reads[Foo]
  }
}
```

SethTisue:
@jtjeferreira interested in returning to this?

jtjeferreira:
@SethTisue I think is better to close it. I wont have time soon :(

PR #594, opened 2021-04-10
sbt 1.5.0

sullis:
# Pull Request Checklist

* [x] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [x] Have you signed the [Lightbend CLA](https://www.lightbend.com/contribute/cla)?
* [x] Have you [squashed your commits](https://www.playframework.com/documentation/latest/WorkingWithGit#Squashing-commits)?
* [ ] Have you added copyright headers to new files?
* [ ] Have you updated the documentation?
* [ ] Have you added tests for any changed functionality?

PR #597, opened 2021-04-15
simplify ScalaTest usage

SethTisue:
~ScalaTest was not fully published for RC1 (Scala.js artifacts were missing), so that's the reason to drop RC1. it's fine, we'd have dropped it soon anyway when adding RC3~ dropped in #602


SethTisue:
```
[error] -- Error: /Users/tisue/play-json/play-json/shared/src/test/scala/play/api/libs/json/JsonSharedSpec.scala:137:6 
[error] 137 |      "js.toJsObject(1)" mustNot typeCheck
[error]     |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[error]     |      The 'mustNot typeCheck' syntax only works with String literals.
```

over at #594, @griggt said he'd look into it

SethTisue:
the Scala 2 jobs are now green

but presumably there's no way to make this PR mergeable until there's an upstream fix (in the ScalaTest repo) for this, as seen in the Scala-3-only failure:

```
[error] -- Error: /home/travis/build/playframework/play-json/play-json/shared/src/test/scala/play/api/libs/json/JsonSharedSpec.scala:137:32 
276[error] 137 |      "js.toJsObject(1)".mustNot(typeCheck)
277[error]     |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
278[error]     |      The 'mustNot typeCheck' syntax only works with String literals.
279[error]     | This location contains code that was inlined from JsonSharedSpec.scala:137
```

over at #594, @griggt said he'd look into it

griggt:
In short, the ScalaTest `typeCheck` matcher is currently broken in Scala 3.

This appears to be known to the maintainers, because the entire test suite for this feature [has been excluded in the Dotty build](https://github.com/scalatest/scalatest/blob/4c7f2aafaaa150a00fd0a83e32e311668da5fbec/project/GenScalaTestDotty.scala#L400).

The immediate issue causing the confusing `only works with String literals` error seen above is that one of the macros contains obsolete/unmaintained code [matching on the wrong tree shape](https://github.com/scalatest/scalatest/blob/4c7f2aafaaa150a00fd0a83e32e311668da5fbec/dotty/matchers-core/src/main/scala/org/scalatest/matchers/CompileMacro.scala#L87-L113): it is expecting an `Apply` node when what we have is a `Literal(StringConstant(_))`.

That is an easy fix, and making that change would allow the tests here to pass. However, the remainder of the current ScalaTest implementation does not conform to the [documented semantics](https://www.scalatest.org/scaladoc/3.2.7/org/scalatest/matchers/must/Matchers.html#checkingThatCodeDoesNotCompile) for the `typeCheck` matcher, namely that `mustNot(typeCheck)` should enforce that the code is free of syntax errors but fails type checking. Instead it would behave identically to `mustNot(compile)`, where either parser or typer errors result in the test succeeding. There are some additional issues (with multiline string literals and `stripMargin`, etc.) that I will omit here for brevity.

Incidentally, the Scala 3 implementation of `illTyped` here has the same issue. Both implementations use the `scala.compiletime.testing.typeChecks` API, which returns `false` on any parser or typer errors.

#### Some options for this PR

 1. Keep `illTyped` as-is (at least for now) but proceed with other changes
 2. Wait for an upstream fix in ScalaTest
 3. Use `mustNot(compile)` instead of `mustNot(typeCheck)`

Regardless of the decision taken here, I plan to work on a PR for ScalaTest to fix `mustNot(typeCheck)` in Scala 3.

griggt:
Upstream PR: https://github.com/scalatest/scalatest/pull/2008

SethTisue:
current status: that PR is now merged, but there is no 3.2.9 yet; perhaps they're waiting for the next Scala 3 release (3.0.0-RC4 or 3.0.0 final)

PR #598, opened 2021-04-15
rename master to main, matching new GitHub default

SethTisue:
hey, other maintainers: I don't know how intertwined the various playframework/* repos are — is there anything else somewhere that would need to be updated?

note that I have not actually renamed the branch yet — I'll go ahead with merging this and renaming on Monday, unless someone tells me I shouldn't


SethTisue:
branch renamed.

PR #633, opened 2021-06-23
Upgrade to jackson 2.12.3

### LINKED ISSUE #632: Upgrade to Jackson 2.12.x
Jackson 2.12.0 was released in November 2020 (latest version is 2.12.3). Would it be possible to upgrade play-json before the 2.10.0 GA relase?

gheine:
Fixes #632

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,10 +1,6 @@
 pullRequests.frequency = "@weekly"
 
 updates.pin  = [
-  // Play JSON should follow the Jackson version in Akka to avoid conflicts in Playframework
-  // https://github.com/akka/akka/blob/master/project/Dependencies.scala#L24
-  { groupId = "com.fasterxml.jackson.core", version = "2.11." }
-  { groupId = "com.fasterxml.jackson.datatype", version = "2.11." }
 ]
 
 commits.message = "${artifactName} ${nextVersion} (was ${currentVersion})"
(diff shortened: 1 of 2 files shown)


ndeverge:
It would be great to merge this one in order to fix CVE-2020-36518.

Thanks

mkurz:
I am interested what the akka team will do, because they are stuck on Jackson 2.11... I am tracking https://github.com/akka/akka/issues/31282 and https://github.com/akka/akka/pull/31281 (which upgrades Jackson only when on Scala 3 however).
If akka don't upgrade, we might finally consider upgrading it for Play though, even in 2.8.x, which should not break akka (but maybe apps?), see https://github.com/akka/akka/issues/31097#issuecomment-1024872663:
> jackson changed between 2.11.4 and 2.13.1. While we don't rely on any of the parts that changed, so Akka works fine with 2.13.1, there could be projects using Akka that do rely on the 2.11 behavior. For that reason we're reluctant to update jackson in Akka, since that might force projects to also update, which might not be easy.

mkurz:
Please have a look my comment in #740, for the next major release IMHO it now makes sense to upgrade to latet Jackson for play-json and play itself.

mkurz:
For CVE-2020-36518 however I think users should upgrade the dependency themselves... Since I don't want to introduce breaking changes into production apps... So everyone has to test carefully. I will write an announcement about the CVE-2020-36518 in the GitHub Discussions soon.

mkurz:
@ndeverge Please see https://github.com/playframework/playframework/discussions/11222 regarding CVE-2020-36518.

ndeverge:
Thanks @mkurz, it's perfectly clear!

PR #634, opened 2021-06-23
fix CI (by only running MiMa for JVM artifacts)

SethTisue:
Currently all PRs in this repo are failing PR validation. This should fix it.


SethTisue:
I misunderstood the cause of the failures. I think it's because 2.10.0-RC3 and 2.10.0-RC4 but never made it to Maven Central. I'm going to delete those tags instead.

Issue #612, opened 2021-05-13
Publish for Scala 3.0.0


SethTisue:
* [x] fix PR validation builds
* [x] merge outstanding PRs
* [x] choose version number
  * 2.10.0-RC5 — I deleted the RC3 and RC4 tags since they didn't publish, but I'm reluctant to reuse even an unreleased version number, so let's call it RC5
* [x] fix publishing (need help from @octonato or somebody else at Lightbend)
* [x] tag it
* [x] wait for release job
  * https://travis-ci.com/github/playframework/play-json/builds/232868402
* [x] wait for artifacts to hit Maven Central
  * https://repo1.maven.org/maven2/com/typesafe/play/play-json_3/2.10.0-RC5/

SethTisue:
status: as discussed at #600, I need help from @octonato or somebody else at Lightbend before I can pull publishing levers

SethTisue:
@marcospereira hey Marcos, Renato is unavailable for the next two weeks. is there any possibility you could step in here...?

xuwei-k:
any updates? 👀

SethTisue:
Marcos and I now have an appointment on Thursday to try and straighten this out.

SethTisue:
Just to recap where things stand, the last publishing effort was https://travis-ci.com/github/playframework/play-json/jobs/503995463 which failed with `2021-05-25 11:35:25.673Z error [Sonatype] [MISSING_PROFILE] Profile com.typesafe.play is not found. Check your sonatypeProfileName setting in build.sbt  - (Sonatype.scala:425)
1295`

SethTisue:
current hypothesis is that #642 will fix it

SethTisue:
2.10.0-RC5 is published. Please report any problems!

PR #642, opened 2021-07-15
Use the correct value for sonatypeProfileName

marcospereira:
`com.typesafe` is the correct value for Play projects. This re-adds the setting
that was before defined in interplay.

References #612.


SethTisue (approved):

SethTisue:
I added a commit which fixes it so the `release` (or `publish`, we weren't using the same name consistently) stage doesn't happen during PR validation.

Issue #186, opened 2018-10-12
Denial of service when parsing JSON object with keys that have the same hash code


plokhotnyuk:
### Play JSON Version (2.5.x / etc)

2.7.0-M1

### API (Scala / Java / Neither / Both)

Scala 2.12.7

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)

Ubuntu 16.04

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

Oracle JDK 11

### Library Dependencies

none

### Expected Behavior

_Sub-linear_ decreasing of throughput when number of JSON object fields (with keys that have the same hash code) is increasing

### Actual Behavior

_Sub-quadratic_ decreasing of throughput when number of JSON object fields (with keys that have the same hash code) is increasing

On contemporary CPUs parsing of such JSON object (with a sequence of 100000 fields like below that is ~1.4Mb) can took more than 100 seconds:

```
{
"!!sjyehe":"",
"!!sjyeiF":"",
"!!sjyej'":"",
"!!sjyfIe":"",
"!!sjyfJF":"",
...
}
``` 

Below are results of benchmark (where `size` is a number of such fields) for different JSON parser including Play-JSON:

```
[info] REMEMBER: The numbers below are just data. To gain reusable insights, you need to follow up on
[info] why the numbers are the way they are. Use profilers (see -prof, -lprof), design factorial
[info] experiments, perform baseline and negative tests that provide experimental control, make sure
[info] the benchmarking environment is safe on JVM/OS/HW level, ask for reviews from the domain experts.
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                                    (size)   Mode  Cnt         Score        Error  Units
[info] ExtractFieldsBenchmark.readAVSystemGenCodec       1  thrpt    7   4725178.602 ±  34049.473  ops/s
[info] ExtractFieldsBenchmark.readAVSystemGenCodec      10  thrpt    7   1271820.062 ±  25118.614  ops/s
[info] ExtractFieldsBenchmark.readAVSystemGenCodec     100  thrpt    7    189164.567 ±   4201.388  ops/s
[info] ExtractFieldsBenchmark.readAVSystemGenCodec    1000  thrpt    7     23129.424 ±    356.359  ops/s
[info] ExtractFieldsBenchmark.readAVSystemGenCodec   10000  thrpt    7      2284.385 ±     55.511  ops/s
[...]

PR #675, opened 2021-11-22
Change `jsObject.value` to `ImmutableLinkedHashMap` to mitigate hash collisions

htmldoug:
After #674, `jsObject.value` would never be used internally (which is great for heap usage and to avoid rehashing), but it's still vulnerable to #186. This PR switches to `ImmutableLinkedHashMap`, an immutable wrapper around `java.util.LinkedHashMap`.

Before:
```
[info] Benchmark                                          (n)   Mode  Cnt    Score    Error  Units
[info] JsonParsing_01_ParseManyFields.parseObjectValue  10000  thrpt    3    1.452 ±  0.144  ops/s
```

After:
```
[info] Benchmark                                          (n)   Mode  Cnt    Score   Error  Units
[info] JsonParsing_01_ParseManyFields.parseObjectValue  10000  thrpt    3  199.307 ± 6.272  ops/s```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -234,17 +234,25 @@ lazy val `play-json` = crossProject(JVMPlatform, JSPlatform)
 
 lazy val `play-jsonJS` = `play-json`.js
 
-lazy val `play-jsonJVM` = `play-json`.jvm.settings(
-  libraryDependencies ++=
-    jacksons ++ {
-      if (isScala3.value)
-        specs2(scalaVersion.value).map(_.exclude("org.scala-lang.modules", "scala-xml_2.13"))
-      else
-        specs2(scalaVersion.value)
-    } :+ (
-      "ch.qos.logback" % "logback-classic" % "1.2.10" % Test
-    ),
-  Test / unmanagedSourceDirectories ++= (docsP / PlayDocsKeys.scalaManualSourceDirectories).value,
+lazy val `play-jsonJVM` = `play-json`.jvm
+  .settings(
+    libraryDependencies ++=
+      jacksons ++ {
+        if (isScala3.value)
+          specs2(scalaVersion.value).map(_.exclude("org.scala-lang.modules", "scala-xml_2.13"))
+        else
+          specs2(scalaVersion.value)
+      } :+ (
+        "ch.qos.logback" % "logback-classic" % "1.2.10" % Test
+      ),
+    Test / unmanagedSourceDirectories ++= (docsP / PlayDocsKeys.scalaManualSourceDirectories).value,
+  )
+  .settings(enableJol)
+
+def enableJol = Seq(
+  libraryDependencies += "org.openjdk.jol" % "jol-core" % "0.16" % Test,
+  Test / javaOptions += "-Djdk.attach.allowAttachSelf",
+  compileOrder := CompileOrder.JavaThenScala,
 )
 
 lazy val `play-json-joda` = project
--- a/play-json/shared/src/main/scala-2.13+/play/api/libs/json/ImmutableLinkedHashMap.scala
+++ b/play-json/shared/src/main/scala-2.13+/play/api/libs/json/ImmutableLinkedHashMap.scala
@@ -1,0 +1,87 @@
+/*
+ * Copyright (C) 2009-2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.api.libs.json
+
+import java.util.{ LinkedHashMap => JLinkedHashMap }
+import scala.collection.AbstractIterator
+import scala.collection.MapFactory
+import scala.collection.immutable.AbstractMap
+import scala.collection.mutable
+
+/**
+ * Wraps a Java LinkedHashMap as a Scala immutable.Map.
+ */
+private[json] class ImmutableLinkedHashMap[A, +B](underlying: JLinkedHashMap[A, B]) extends AbstractMap[A, B] {
+
+  override def get(key: A): Option[B] = Option(underlying.get(key))
+
+  override def removed(key: A): Map[A, B] = {
+    val c = shallowCopy()
+    c.remove(key)
+    new ImmutableLinkedHashMap(c)
+  }
+
+  override def updated[V1 >: B](key: A, value: V1): Map[A, V1] = {
+    val c = shallowCopy[V1](size + 1)
+    c.put(key, value)
+    new ImmutableLinkedHashMap(c)
+  }
+
+  override def mapFactory: MapFactory[Map] = ImmutableLinkedHashMap
+
+  override def iterator: Iterator[(A, B)] = new AbstractIterator[(A, B)] {
+    private[this] val ui = underlying.entrySet().iterator()
+
+    override def hasNext: Boolean = ui.hasNext
+
+    override def knownSize: Int = if (underlying.isEmpty) 0 else super.knownSize
+
+    override def next(): (A, B) = {
+      val e = ui.next()
+      (e.getKey, e.getValue)
+    }
+  }
+
+  override def knownSize: Int = underlying.size()
+  override def size: Int      = underlying.size()
+
+  private def shallowCopy[V1 >: B](sizeHint: Int = size): JLinkedHashMap[A, V1] = {
+    val c = new JLinkedHashMap[A, V1](sizeHint)
+    for ((k, v) <- this) c.put(k, v)
+    c
+  }
+}
+
+private[json] object ImmutableLinkedHashMap extends MapFactory[Map] {
+  private object EmptyMap extends ImmutableLinkedHashMap[Any, Nothing](new JLinkedHashMap(0))
+
+  override def empty[K, V]: Map[K, V] = EmptyMap.asInstanceOf[Map[K, V]]
+
+  override def from[K, V](it: IterableOnce[(K, V)]): Map[K, V] = (newBuilder ++= it).result()
+
+  override def newBuilder[A, B]: mutable.Builder[(A, B), Map[A, B]] = new mutable.Builder[(A, B), Map[A, B]] {
+    private[this] var lhm = new JLinkedHashMap[A, B](0)
+
+    override def clear(): Unit = lhm.clear()
+
+    override def sizeHint(size: Int): Unit = if (size > 0 && lhm.isEmpty) lhm = new JLinkedHashMap[A, B](size)
+
+    override def result(): Map[A, B] = {
+      if (lhm.isEmpty) empty
+      else new ImmutableLinkedHashMap(lhm)
+    }
+
+    override def addOne(elem: (A, B)): this.type = {
+      lhm.put(elem._1, elem._2)
+      this
+    }
+
+    override def addAll(xs: IterableOnce[(A, B)]): this.type = {
+      sizeHint(xs.knownSize)
+      xs.iterator.foreach(addOne)
+      this
+    }
+  }
+}
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala
@@ -136,7 +136,7 @@ case class JsObject(
    */
   lazy val value: Map[String, JsValue] = underlying match {
     case m: immutable.Map[String, JsValue] => m
-    case m                                 => m.toMap
+    case m                                 => JsObject.createFieldsMap(m)
   }
 
   /**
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala [hunk 1]
```diff
@@ -136,7 +137,7 @@ case class JsObject(
    */
   lazy val value: Map[String, JsValue] = underlying match {
     case m: immutable.Map[String, JsValue] => m
-    case m                                 => m.toMap
+    case m                                 => (TreeMap.newBuilder[String, JsValue] ++= m).result()
```
htmldoug, line 140: On second thought, can we just return `m` directly?

```suggestion
    case m                                 => m
```

The scaladoc says "The value of this JsObject as an _immutable_ map", but the type signature is only `scala.collection.Map`. Anyone requiring a `scala.collection.immutable.Map` would need to call `.toMap` anyway or do an `.asInstanceOf[s.c.i.Map]`.

If we just return `m` directly and change remove "immutable" from the scaladoc, it would do good things for memory footprint and throughput:
```
[info] Benchmark                                          (n)   Mode  Cnt    Score    Error  Units
[info] JsonParsing_01_ParseManyFields.parseObjectValue  10000  thrpt    3  167.452 ± 37.356  ops/s
```

This would shift the responsibility for handling hashCode collisions onto users, but this method isn't suggested usage anyway.

@gmethvin, WDYT?

htmldoug: I've created a separate PR for that option: #676.

gmethvin: I don't think it's worth the breaking change. Since the method isn't suggested usage I don't think users will benefit much from the change anyway.

htmldoug: Looks like `.value` is [getting used](https://github.com/julienrf/play-json-derived-codecs/blob/aa0c78a008e088dbd77768e04a3e633b48f8f24e/library/src/main/scala/julienrf/json/derived/DerivedOWrites.scala#L97) in the wild.

I'll reopen this PR to keep both options on the table.

htmldoug: > not worth the breaking change

I'm having difficultly thinking of a scenario where it would be breaking. Method signatures wouldn't change. Anyone doing `.value.asInstanceOf[s.c.i.Map]` should know that's unsound since `JsObject` is non-final and it's already possible to subclass this and make `value` return other representations. Are there other scenarios to be concerned about?

gmethvin: yeah, I wouldn't say it's _technically_ a breaking change, but Scala users typically want immutable objects so we should plan for the immutable case.

htmldoug: Hmm, okay. How about making `underlying` immutable via a `class ImmutableJMapWrapper[K, V](jmap: java.util.Map[K, V]) extends scala.collection.immutable.Map[K, V]` instead?

htmldoug:
```scala
/**
  * Wraps a Java String-keyed map as a Scala one with fallback to TreeMap for modifications.
  */
private[json] class ImmutableJMapWrapper[V](underlying: java.util.Map[String, V]) extends scala.collection.immutable.Map[String, V] {

  override def removed(key: String): Map[String, V] = to(TreeMap).removed(key)

  override def updated[V1 >: V](key: String, value: V1): Map[String, V1] = to(TreeMap).updated(key, value)

  override def get(key: String): Option[V] = Option(underlying.get(key))

  override def iterator: Iterator[(String, V)] = new AbstractIterator[(String, V)] {
    private[this] val ui = underlying.entrySet().iterator()

    override def hasNext: Boolean = ui.hasNext

    override def knownSize: Int = if (underlying.isEmpty) 0 else super.knownSize

    override def next(): (String, V) = {
      val e = ui.next()
      (e.getKey, e.getValue)
    }
  }
}
```

Benefits:
- immutable
- replaces JMapWrapper, so same memory footprint.
- no copy on `value`
- third party libraries like play-json-derived-codecs gain access to the same performance that the play-json macros enjoy by using the private `underlying` field

Slightly weird:
- no way to `clone()` the java Map, so modifications need to copy to some other type (AFAIK, only TreeMap is safe against #186)
- `TreeMap` requires an `Ordering`. I just narrowed key to `String` to avoid falling into the next object alignment. `String` is all we need anyway.
- we'd need a separate implementation for scala-2.12-, but it's only 10 lines.

WDYT?

gmethvin: How does the performance of `LinkedHashMap` compare to `immutable.TreeMap` in general? I wonder if we should just change the underlying map to `immutable.TreeMap` and avoid using the `LinkedHashMap` hack altogether...

gmethvin: Oh, I forgot something potentially important...

IIRC, we use `LinkedHashMap` instead of `HashMap` because some users are expecting a specific ordering of keys. Technically the ordering of keys in JSON isn't significant, but it's something we might want to be careful about breaking.

gmethvin: I guess using the `ImmutableJMapWrapper` as the `underlying` map could be a good idea. It feels like an ugly hack but the whole `LinkedHashMap` thing feels that way too.

gmethvin: All that said, the "correct" solution here is to fix the bug in Scala, so what we're doing here is hopefully just a "temporary" workaround.

htmldoug: > How does the performance of LinkedHashMap compare to immutable.TreeMap in general?

FWIW, `j.u.LinkedHashMap` had 1.3x-1.4x of the memory footprint of `s.c.i.TreeMap` in my testing, but has faster lookup over the common case from the open addressing hash table.

As for `.value`, all of the `s.c.immutable.Map` implementations (except `Map1-4`) destroy order in one way or another on `.toMap`. Our `ImmutableJMapWrapper` would be an exception by preserving order--another benefit.

> All that said, the "correct" solution here is to fix the bug in Scala

That's a good point, although I haven't heard any developments since https://github.com/scala/bug/issues/11203 was marked `won't fix`. I know https://github.com/scala/scala/pull/7633 is available, but that's mutable and doesn't preserve order. 

> It feels like an ugly hack

Well, the [MapAndVector](https://github.com/circe/circe/blob/e6685317bc9e6e65539bb6194900e0e754fbcb7c/modules/core/shared/src/main/scala/io/circe/JsonObject.scala#L384-L387) approach is cleaner, but would be breaking to apply here for anyone [using the case class `unapply`](https://github.com/com-lihaoyi/upickle/blob/106a706287c3f3cfe615cf5c65e692ed81181b87/ujson/play/src/ujson/play/PlayJson.scala#L15) to get at the `underlying` map--which I only just realized was possible.

It's worth mention that I've been experimenting with applying V8's hidden classes to `JsObject` to cut memory footprint. (I love that I can stuff any Map I want into it!) Gains are quite promising for large messages. I'm aware that's probably more crazy than you'd want in the base implementation, but regardless, here are the bones for potential inspiration. It's ordered, as well.

```
class ShapeMap[V](
  values: Array[V],
  sharedShape: Map[String, Int], // array indexes (j.u.HashMap)
  sharedShapes: ConcurrentMap[Seq[String], Map[String, Int]]
) extends scala.collection.immutable.Map[String, V]
```

Anyway, I'm just here to upstream any useful low hanging fruit I'm finding while benching and poking around in the weeds. Do you want me to go ahead with a PR for `ImmutableJMapWrapper`?

gmethvin: `ImmutableJMapWrapper` seems like an ok solution.

Honestly it would be really nice if we could just break the ordering and use `TreeMap` as the underlying data structure, but I don't remember exactly what the concerns were there.

htmldoug: Happy new year!

I've rewritten the PR to make an immutable wrapper around `java.util.LinkedHashMap`. I coupled directly to LHM rather than all `j.u.Map` since it's easier to support `+` and `++`. Generality seems unnecessary.

I've explored the `Map` space further and ended up using the following implementation (along with specialized Map1-4) in my own project:
```
class SortedArraysMap[V](
  private val sortedKeys: Array[String], // sorted, binary search lookup
  private val vals: Array[V]             // parallel to sortedKeys
) extends scala.collection.immutable.Map[String, V]
```
Everything's mostly `log(n)` (like `TreeMap`), but the footprint is much smaller.
Map memory footprint by cardinality (1)

I've relaxed `JsObject.+`, etc. to delegate to the underlying `Map` implementation and avoid forcing `j.u.LHM`. If users don't care about order and want `log(n)` `+`, `++`, etc. instead of `n` with LHM, they have the option of using `TreeMap`, `s.c.i.HashMap`, or something like my `SortedArraysMap`.

Please review when you get a chance. Criticism welcome.

play-json/shared/src/main/scala-2.13+/play/api/libs/json/ImmutableLinkedHashMap.scala
```diff
@@ -0,0 +1,87 @@
...
+import java.util.{ LinkedHashMap => JLinkedHashMap }
+import scala.collection.AbstractIterator
+import scala.collection.MapFactory
+import scala.collection.immutable.AbstractMap
+import scala.collection.mutable
+
+/**
+ * Wraps a Java LinkedHashMap as a Scala immutable.Map.
+ */
+private[json] class ImmutableLinkedHashMap[A, +B](underlying: JLinkedHashMap[A, B]) extends AbstractMap[A, B] {
```
gmethvin, line 16: Maybe this should be called `UnmodifiableLinkedHashMap` since it doesn't guarantee that other code can't change the underlying collection.

gmethvin (commented):
Overall I'm pretty happy with this. Maybe @cchantep should take a look as well?

play-json/shared/src/main/scala-2.13+/play/api/libs/json/ImmutableLinkedHashMap.scala
htmldoug: Subtle distinction. The java `Unmodifiable` classes allow users to wrap mutable collections they control whereas the immutable ones don't (with the exception of `scala.collection.immtuable.ArraySeq.unsafeWrapArray`, `akka.util.ByteString.fromArrayUnsafe`).

This `ImmutableLinkedHashMap` package-protects the constructor (and the companion object). This seems at least as strong as the others. Technically, the user can use reflection to break any guarantees if they really want to.

I think it all comes down to intent. I read "unmodifiable" as "you can't modify it, but I can and might", and immutable as "I'm not gonna modify it either".

htmldoug:
```scala
val arr = Array[AnyRef]("Vector is immutable, right?")
val as = ArraySeq.unsafeWrapArray(arr)
val v = Vector.from(as)
println(v) // Vector(Vector is immutable, right?)
arr(0) = "This makes me uncomfortable"
println(v) // Vector(This makes me uncomfortable)
```

gmethvin: yeah, it's an internal class so I'm not super concerned about it either way 🤷

build.sbt
```diff
@@ -253,17 +253,25 @@ lazy val `play-json` = crossProject(JVMPlatform, JSPlatform)
...
+          specs2(scalaVersion.value)
+      } :+ (
+        "ch.qos.logback" % "logback-classic" % "1.2.10" % Test
+      ),
+    Test / unmanagedSourceDirectories ++= (docsP / PlayDocsKeys.scalaManualSourceDirectories).value,
+  )
+  .settings(enableJol)
+
+def enableJol = Seq(
+  libraryDependencies += "org.openjdk.jol" % "jol-core" % "0.16",
```
ihostage, line 272: I have doubts about this dependency 🤔 
Do we need this in runtime? Looks like that is needed only for tests.

htmldoug: Good catch! Moved to `Test`.

ihostage (approved):
LGTM.
@gmethvin @cchantep WDYT?

gmethvin:
LGTM

mkurz:
Just to let you know, the added memory footprint spec [failed](https://github.com/playframework/play-json/actions/runs/4881181978/jobs/8709892757) when upgrading to Scala 3.2.1+, because with that scala version memory usage decreased. See #848 where I adjusted the expected values and for further explanation.

PR #679, opened 2021-11-29
[main] update sonatype profile name

octonato:


build.sbt
```diff
@@ -137,8 +137,7 @@ lazy val commonSettings = Def.settings(
   Compile / doc / scalacOptions ++= Seq(
     // Work around 2.12 bug which prevents javadoc in nested java classes from compiling.
     "-no-java-comments",
-  ),
-  sonatypeProfileName := "com.typesafe"
```
mkurz, line 141: @octonato You removed it, but shouldn't it be set to `com.typesafe.play`?

mkurz (changes requested):

build.sbt
octonato: No, that's correct. 

The main branch is not using Interplay any more and therefore, if you say nothing, the profile name will be the same as the groupId, so `com.typesafe.play`. 

When we removed Interplay we had to force it to be `com.typesafe`, but now we want it to be the groupId. 

In the [2.8.x branch](https://github.com/playframework/play-json/pull/678/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R128), is the other way around. There we use Interplay and it sets it to `com.typesafe` and we need to force it to `com.typesafe.play`. 

Confusing, no?

mkurz (approved):

PR #692, opened 2022-01-09
Optimize `JsObject.equals()`/`hashCode()`

htmldoug:
## Summary
For `JsObject.equals()`/`hashCode()`:
- improves runtime by avoiding making two copies of the underlying data
- improves memory footprint by avoiding holding `lazy val fields: collection.Seq[(String, JsValue)]` after the computation is performed.
- Benchmark: [jmh results](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/htmldoug/6ebb5f3f169e35535a99e2d198c9153d/raw/f8f1a3b905de98829e607ad41fe8ffa936ffd6c8/report-before.json,https://gist.githubusercontent.com/htmldoug/6ebb5f3f169e35535a99e2d198c9153d/raw/f8f1a3b905de98829e607ad41fe8ffa936ffd6c8/report-after.json)

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala
@@ -5,6 +5,7 @@
 package play.api.libs.json
 
 import scala.collection._
+import scala.util.hashing.MurmurHash3
 
 case class JsResultException(errors: collection.Seq[(JsPath, collection.Seq[JsonValidationError])])
     extends RuntimeException(s"JsResultException(errors:$errors)")
@@ -188,14 +189,20 @@ case class JsObject(
     merge(this, other)
   }
 
-  override def equals(other: Any): Boolean = other match {
-    case that @ JsObject(_) => (that.canEqual(this)) && fieldSet == that.fieldSet
-    case _                  => false
+  override def equals(other: Any): Boolean = {
+    other match {
+      case o: AnyRef if this.eq(o) =>
+        true
+      case JsObject(that) =>
+        underlying == that
+      case _ =>
+        false
+    }
   }
 
   def canEqual(other: Any): Boolean = other.isInstanceOf[JsObject]
 
-  override def hashCode: Int = fieldSet.hashCode()
+  override def hashCode(): Int = MurmurHash3.unorderedHash(underlying, MurmurHash3.setSeed)
 }
 
 object JsObject extends (Seq[(String, JsValue)] => JsObject) {
--- a/benchmarks/src/main/scala/play/api/libs/json/JsObjectBench.scala
+++ b/benchmarks/src/main/scala/play/api/libs/json/JsObjectBench.scala
@@ -1,0 +1,56 @@
+package play.api.libs.json
+
+import org.openjdk.jmh.annotations._
+
+import java.util.concurrent.TimeUnit
+
+/**
+ * ==Quick Run==
+ * benchmarks / Jmh / run .*JsObjectBench
+ *
+ * ==Profile with Flight Recorder==
+ * benchmarks / Jmh / run -prof jfr .*JsObjectBench
+ *
+ * ==Jmh Visualizer Report==
+ * benchmarks / Jmh / run -f5 -prof gc -rf json -rff JsObjectBench-results.json .*JsObjectBench
+ *
+ * ==Sample Results==
+ * {{{
+ * Benchmark                    (size)   Mode  Cnt       Score       Error   Units
+ * JsObjectBench.oEqualsCopy        15  thrpt    5    4386.757 ±   124.675  ops/ms
+ * JsObjectBench.oEqualsEq          15  thrpt    5  430769.750 ± 43891.912  ops/ms
+ * JsObjectBench.oEqualsNe          15  thrpt    5  321636.207 ± 23899.623  ops/ms
+ * JsObjectBench.oHashCodeWarm      15  thrpt    5    4024.849 ±   276.276  ops/ms
+ * }}}
+ *
+ * @see https://github.com/ktoso/sbt-jmh
+ */
+@Warmup(iterations = 20, time = 1, timeUnit = TimeUnit.SECONDS)
+@Measurement(iterations = 10, time = 1, timeUnit = TimeUnit.SECONDS)
+@State(Scope.Benchmark)
+@BenchmarkMode(Array(Mode.Throughput))
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Fork(
+  jvmArgsAppend =
+    Array("-Xmx350m", "-XX:+HeapDumpOnOutOfMemoryError", "-XX:-BackgroundCompilation", "-XX:-TieredCompilation"),
+  value = 1
+)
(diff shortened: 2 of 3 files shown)


play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala
```diff
@@ -190,14 +191,20 @@ case class JsObject(
     merge(this, other)
   }
 
-  override def equals(other: Any): Boolean = other match {
-    case that @ JsObject(_) => (that.canEqual(this)) && fieldSet == that.fieldSet
-    case _                  => false
+  override def equals(other: Any): Boolean = {
+    other match {
+      case o: AnyRef if this.eq(o) =>
```
cchantep, line 196: How could this happen?

htmldoug: Good question! As an illustrative example, consider a json text that looks like `[[{"k":0}], [{"k":0}],[{"k":0}],[{"k":0}],...]` whose length approaches the jvm's total heap size. `Json.parse(file)` would throw `OutOfMemoryError`, but is it _possible_ to represent such a text as a `JsValue`?

Currently, jackson-core will return the same `"k"` `String` instance for all keys in the example, but that's only a small part of the eventual `JsValue` memory footprint. It's possible to apply the same technique to the other duplicated values as well.

In weepickle terms, a minimal implementation could look like:
```scala
val jsValue = FromJson(inputStream).transform(new DdVisitor(PlayJson))

class DdVisitor[T, J](underlying: Visitor[T, J])
 (implicit cache: mutable.Map[Any, Any] = mutable.Map.empty)
 extends Visitor.Delegate[T, J](underlying)
   with RecursiveDelegate[T, J] {

 override def asRecursiveDelegate[A, B](visitor: Visitor[A, B]) =
   new DdVisitor(
     visitor.map { jsValue =>
       cache
         .getOrElseUpdate(jsValue, jsValue)
         .asInstanceOf[B]
     }
   )
}
```

As we walk the input depth-first, every `JsValue` node is run through a `Map[JsValue, JsValue]` to return a canonical instance. Each `{"k":0}` would result in a cache hit and return a single canonical instance.

Back to your question, this line would be hit when we attempt to deduplicate `[{"k":0}]`. In order to check if two `JsArray` are identical, it needs to test that all elements pass equals(). Since the `{"k":0}` `JsObject` that's present in both arrays would have already been canonicalized to the same instance, `this eq other` would be true for that comparison.

I wish this were a purely hypothetical scenario. In practice, we talk to some APIs that return duplication-heavy JSON texts that have OOME'd our apps before. My favorite example was a 5.4 GB `JsObject` in heap whose UTF-8 input compressed down to 1.7 MB. Those units are not typos. That's ~3,000x memory amplification.

One of the things we like about play-json is that it gives us the flexibility to mitigate these problems by extending the JsValue classes. Applying deduplication brought the 5.4 GB JsValue down to a 205 MB JsValue. Bringing our [own Map/Seq](https://github.com/playframework/play-json/pull/675#discussion_r780752661) and doing some unboxing got the same input down to a 23 MB JsValue. Throughput was also faster due to eliminated GC pressure. Win-win. (We also logged an issue with the service responding with the extra data to please stop sending us info we don't need, but I was on-call and wanted to stop getting paged about this.)

Some APIs are highly normalized and duplicate-free, but in practice, I suspect many of them contain significant duplication. Using the few thousand test json files in our test double service as a proxy for the real-world data we deal with, I found that 71% of the nodes were present somewhere else in the same json text. Duplication wasn't present only in primitives either. I found significant duplication up to height=5 of the trees. I'm very curious if we just deal with bad REST APIs or if this is more widespread.

Anyway, over typical usage this line won't hit, although it's quite cheap, especially compared to the previous implementation that constructed two copies of `underlying` on equals()--one of them a `HashSet`! If needed, we could remove it and get by with the underlying Map performing the `eq` check, but while I was in here, I saw the opportunity to short-circuit, tried it while benchmarking, and left it.

Thanks for the early feedback on this draft! Code is actually ready for review, but I didn't want to send reviewer notifications until I'd written the PR description. Now that I've typed all this out, maybe I'll just skip it. :)

cchantep: K, would be interested to add a benchmark after.

htmldoug: Before:
```
[info] Benchmark                    (size)   Mode  Cnt  Score   Error   Units
[info] JsObjectBench.oEq                 1  thrpt    5  5.221 ± 0.194  ops/us
[info] JsObjectBench.oEq                15  thrpt    5  0.292 ± 0.007  ops/us
[info] JsObjectBench.oEquals             1  thrpt    5  5.050 ± 0.124  ops/us
[info] JsObjectBench.oEquals            15  thrpt    5  0.279 ± 0.016  ops/us
[info] JsObjectBench.oHashCodeWarm       1  thrpt    5  8.608 ± 0.524  ops/us
[info] JsObjectBench.oHashCodeWarm      15  thrpt    5  0.513 ± 0.012  ops/us
```

After:
```
[info] Benchmark                    (size)   Mode  Cnt    Score    Error   Units
[info] JsObjectBench.oEq                 1  thrpt    5  420.943 ± 13.411  ops/us
[info] JsObjectBench.oEq                15  thrpt    5  418.040 ± 19.643  ops/us
[info] JsObjectBench.oEquals             1  thrpt    5  314.461 ± 14.557  ops/us
[info] JsObjectBench.oEquals            15  thrpt    5  310.431 ± 25.338  ops/us
[info] JsObjectBench.oHashCodeWarm       1  thrpt    5   20.217 ±  1.559  ops/us
[info] JsObjectBench.oHashCodeWarm      15  thrpt    5    3.830 ±  0.311  ops/us
```

htmldoug: Longer run, higher precision, prettier graphs:
https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/htmldoug/6ebb5f3f169e35535a99e2d198c9153d/raw/f8f1a3b905de98829e607ad41fe8ffa936ffd6c8/report-before.json,https://gist.githubusercontent.com/htmldoug/6ebb5f3f169e35535a99e2d198c9153d/raw/f8f1a3b905de98829e607ad41fe8ffa936ffd6c8/report-after.json

gmethvin (approved):
Seems reasonable to me, and a nice performance improvement!

cchantep (approved):
Merge ready?

cchantep:
See throughput bench, if mem footprint bench is a quickwin ...

htmldoug:
Sorry, I don't follow. Is there an existing test you want me to look at?

The memory footprint tests I'm referring to are [here](https://github.com/playframework/play-json/pull/675/files#diff-2ac49b0aeee84afe19f175984d3dc20d55e3cdb827988225f9a749ab9f2e9659R15-R16). The benefits are less about throughput and more about preventing OOME conditions over large or many inputs that have to be in memory at once.

cchantep:
Fine for me

PR #696, opened 2022-01-12
Use regexp for Mergify conditions

ihostage:
More details in [Mergify docs](https://docs.mergify.com/configuration/#regular-expressions).
I checked that in [regex101](https://regex101.com).
Now Mergify should be more loyal for updating patch versions of Scala. 🤞


mkurz (approved):
Great!

PR #702, opened 2022-01-20
perf: Override ImmutableLinkedHashMap.++

htmldoug:
I forgot to implement `++` for 2.13+ in #675. It's already in place for 2.12-. This will avoid creating O(n) `JLinkedHashMap` copies.


BillyAutrey:
I figure you're still in progress here.  Any chance you can add a test for this behavior?

htmldoug:
Thanks for the review! Looking into the test failure made me remember that we don't have to implement ++ here at all in 2.13. `concat()` goes back to the builder which special cases empty objs. I love the new 2.13 collections!

If you're still interested in reviewing, I'd love for eyes on #701, particularly wrt any use of the lazy vals in the wild that I may be unaware of.

PR #705, opened 2022-01-24
Use a reusable workflows

ihostage:


.github/workflows/publish.yml
```diff
@@ -10,26 +10,4 @@ on:
...
-        uses: coursier/cache-action@v6.2
-
-      - name: Publish artifcats
-        run: sbt ci-release
-        env:
-          PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
-          PGP_SECRET: ${{ secrets.PGP_SECRET }}
-          SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
-          SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
+    uses: ihostage/.github/.github/workflows/publish.yml@main
```
ihostage, line 13: 🤔 https://github.community/t/using-organization-secrets-in-reusable-workflows/205207
Need to try without passing secrets explicitly. If it won't work then we'll change it.

ihostage: Nope, it didn't work 😞 Need to set secrets explicitly. Hurtful 😞 
https://github.com/playframework/play-json/runs/4985224723?check_suite_focus=true

mkurz (approved):
Rebased and changed to the correct .github repo.
Let's see if mergify merges.

PR #740, opened 2022-04-05
Jackson 2.13

mkurz:
So akka is very conservative about upgrading jackson, because they do not want to break any existing apps.
However, based on these comments I assume akka itself will should work just fine with latest jackson since it "do[es]n't rely on any of the parts that changed":
* https://github.com/akka/akka/issues/31097#issuecomment-1024872663
* https://github.com/akka/akka/issues/31282#issuecomment-1087439343
* https://github.com/akka/akka/issues/31237#issuecomment-1064288737

So I suggest, if everything works out and CI agrees to finally upgrade jackson in play-json and play itself to the latest version for the next major release.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,8 +21,8 @@ def specs2(scalaVersion: String) =
     ("org.specs2" %% s"specs2-$n" % "4.14.1").cross(CrossVersion.for3Use2_13) % Test
   }
 
-val jacksonVersion         = "2.11.4"
-val jacksonDatabindVersion = jacksonVersion
+val jacksonVersion         = "2.13.2"
+val jacksonDatabindVersion = "2.13.2.2"
 
 val jacksonDatabind = Seq(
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -21,8 +21,8 @@ def specs2(scalaVersion: String) =
     ("org.specs2" %% s"specs2-$n" % "4.14.1").cross(CrossVersion.for3Use2_13) % Test
   }
 
-val jacksonVersion         = "2.11.4"
-val jacksonDatabindVersion = jacksonVersion
+val jacksonVersion         = "2.13.2"
+val jacksonDatabindVersion = "2.13.2.2"
```
mkurz, line 25: Only databind has that `.2` release, core and others don't have that, you can check here:
* https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/
* https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-core/

PR #757, opened 2022-05-03
Switch to cmd.yml

mkurz:
Testing
* https://github.com/playframework/.github/pull/19

Also see
* https://github.com/playframework/playframework/pull/11259 and
* https://github.com/playframework/playframework/pull/11262

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -31,7 +31,7 @@ jobs:
       - "check-code-style"
       - "check-binary-compatibility"
       - "check-docs"
-    uses: playframework/.github/.github/workflows/contact@playframework.com
+    uses: playframework/.github/.github/workflows/cmd.yml@dynamic_matrix
     with:
       java: >-
         [ "11", "8" ]
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -11,7 +11,7 @@ concurrency:
 jobs:
   check-code-style:
     name: Code Style
-    uses: playframework/.github/.github/workflows/cmd.yml@dynamic_matrix
+    uses: playframework/.github/.github/workflows/cmd.yml@cmd-ihostage
     with:
       cmd: sbt validateCode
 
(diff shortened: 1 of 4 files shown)


mkurz:
I have no idea why GitHub action test workflow isn't running for this PR...

BillyAutrey:
> I have no idea why GitHub action test workflow isn't running for this PR...

Found the error here - https://github.com/playframework/play-json/actions/runs/2266633506/workflow#L34

If you click on `Checks` for a PR that doesn't run checks, you can see which step has an error on the left.  It shows up as a text box with an exclamation point.

From there, it says:
> [Invalid workflow file: .github/workflows/build-test.yml#L34](https://github.com/playframework/play-json/actions/runs/2266633506/workflow)
invalid value workflow reference: workflow file should have either a '.yml' or '.yaml' file extension

Clicking in on that error takes me to the first link.  Seems it doesn't like the format of that command file, because it thinks `playframework/.github/.github/workflows/contact@playframework.com` isn't a proper yaml.

.github/workflows/build-test.yml
```diff
@@ -31,12 +31,14 @@ jobs:
       - "check-code-style"
       - "check-binary-compatibility"
       - "check-docs"
-    uses: playframework/.github/.github/workflows/sbt-matrix.yml@v1
+    uses: playframework/.github/.github/workflows/contact@playframework.com
```
BillyAutrey, line 34:
```suggestion
    uses: playframework/.github/.github/workflows/cmd.yml@dynamic_matrix
```

BillyAutrey:
Changed the file, and now it works.  :)

mkurz:
> If you click on `Checks` for a PR that doesn't run checks, you can see which step has an error on the left. It shows up as a text box with an exclamation point.

Actually I do know that and used that error messages a lot when I worked on all that GitHub actions stuff. However yesterday it was different; I had the feeling the check actually did not even run, like the workflow file did not get parsed at all (yet). Like there was not even the check showing up. Maybe there was no free vm from GitHub to pick the workflow or whatever. Anyway, thanks!

ihostage (approved):

PR #761, opened 2022-05-18
GitHub actions last round + badges

mkurz:


.github/workflows/publish.yml
```diff
@@ -4,8 +4,9 @@ on:
   push:
     branches: # Snapshots
       - main
-      - 2.8.x
```
mkurz, line 7: Adding 2.8.x here does not make any sense, it only makes sense in the 2.8.x branch itself.

PR #766, opened 2022-05-23
Scala Steward should ignore Jackson modules except `jackson-core`

ihostage:
Inspired by #763, #764 and #765


mkurz (commented):
👍 😉 (GitHub requests my review still)

PR #793, opened 2022-09-29
Drop Java 8

mkurz:


cchantep:
Hi,

Play-JSON 2.10.0 has just been published, I would have been pleased to upgrade to ... but as working a app that must run on Java 8 ...

We can considered that we have to upgrade ... JVM is a system component, and in some context it's not as easy ... compared to the fact that it has no cost for the Play JSON build ...

mkurz:
@cchantep Hmm... I think it's time to move on at some point? Java 8 was released almost 10 years ago. Some libraries started to drop Java 8 artifacts. Bigger frameworks like Spring, with version 6, even **require** Java 17 (!) already...
I do understand that there will always be organziations which follow a very conservative approach of upgrading software components, but at some point we have to make a cut IMHO.
Is it that hard for you/your organzation to upgrade to Java 11/Java 17? Like are you using some special Java 8 flags/workarounds etc.? Or is it "just" a matter of manpower, time and testing to perform a Java upgrade?

cchantep:
"Or is it "just" a matter of manpower, time and testing to perform a Java upgrade?" your argument can easily to reverted ... as the Java8 compatibility has no cost there.

PR #813, opened 2022-10-17
adjust JsonMemoryFootprintSpec for Scala 2.13.10 as used by Scala 3

SethTisue:
at present, Scala 3 uses the Scala 2.13.8 standard library. I am working on a PR (lampepfl/dotty#16074) to upgrade it to 2.13.10. in CI, play-json failed in the community build after the version bump

here in this PR, I've blindly updated the numbers to make the test pass again, but I don't actually know what the cause for the change might have been. so I'm looking for your confirmation that there isn't some cause for concern here


SethTisue:
@htmldoug `JsonMemoryFootprintSpec` appears to be your brainchild?

SethTisue:
That's good to hear! I'll close this PR for now since there isn't yet a proper Scala 3 release that needs it.

PR #826, opened 2022-11-22
Support Scala Native

lolgab:
# Pull Request Checklist

## Purpose

Add support for Scala Native


play-json/js-native/src/main/scala/StaticBindingJsNative.scala
```diff
@@ -0,0 +1,113 @@
+/*
+ * Copyright (C) 2009-2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.api.libs.json
+
+import java.io.InputStreamReader
+
+private[json] object StaticBindingJsNative {
```
lolgab, line 9: The code in this file was move from `StaticBinding` in the js implementation.

play-json/native/src/main/scala/StaticBinding.scala
```diff
@@ -0,0 +1,58 @@
...
+      case '\b' => "\\b"
+      case '\f' => "\\f"
+      case '\n' => "\\n"
+      case '\r' => "\\r"
+      case '\t' => "\\t"
+      case '\\' => "\\\\"
+      case '\"' => "\\\""
+      case c    => c.toString
+    }
+    val stringified = if (s == null) "null" else s""""${s.flatMap(escaped)}""""
```
cchantep, line 54:
```suggestion

    val stringified = if (s == null) "null" else s.flatMap(escaped)}
    
```

scripts/test-code.sh
```diff
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
 
 echo MATRIX_SCALA=$MATRIX_SCALA
-sbt -DscalaJSStage=full ++$MATRIX_SCALA test publishLocal || exit 1
+sbt -DscalaJSStage=full \
+    -J-XX:MinRAMPercentage=90.0 \
```
cchantep, line 5: Why would it be required?

play-json/native/src/main/scala/StaticBinding.scala
lolgab: I did `s""""${s.flatMap(escaped)}""""` to add an extra pair of quotes around `s.flatMap(escaped)`

scripts/test-code.sh
lolgab: It's because Scala Native requires more RAM to build the binary with the tests.
`-XX:MinRAMPercentage=90.0` and `-XX:MaxRAMPercentage=90.0` allow us to use the RAM available in the machine (especially in the Github Action executor) without making assumptions about its memory size.

play-json/js-native/src/main/scala/StaticBindingJsNative.scala
BillyAutrey: I'm just curious here - what's the point of this restructure?

play-json/native/src/main/scala/StaticBinding.scala
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (C) 2009-2021 Lightbend Inc. <https://www.lightbend.com>
```
BillyAutrey, line 2: I figure this is needed to ensure linting works on the new rules. But this isn't really associated with Lightbend anymore. @mkurz we should visit revamping the requirements on copyrights and file headers.

play-json/js-native/src/main/scala/StaticBindingJsNative.scala
lolgab: > Ah, I think I see - you're making the underlying functionality common, so that both `js` and `native` `StaticBinding` implementations can access this code?

Exactly, to avoid copy-pasting a lot of code, I put the common code here so it's shared and called back and forth from the Native and JS implementations.

BillyAutrey: A small nit then - I found the `js-native` and `StaticBindingJsNative` names a bit confusing. The name doesn't really convey the intent, and isn't intuitive for further possible expansions.

lolgab: the `js-native` directory is automatically added by `sbt-crossproject`.
About `StaticBindingJsNative` I just added the platforms names to the original file name. If you have a better suggestion I change it. I don't have any better name in mind.. maybe `StaticBindingNonJvm` ?

BillyAutrey: Ah, I had a feeling it was sbt plugin shenanigans causing the package name. Makes sense, given that the subprojects aren't explicitly depending on one another in `build.sbt`.

I think it's really a nit though, I hate to delay for this. Let's leave it as is, and leave better naming for later. This isn't a public API, so it shouldn't affect very many users.

I will say though, `StaticBindingNonJvm` or `StaticBindingNonJvmShared` is a bit more descriptive.

BillyAutrey (approved):
Approved, though we should track the comments I made.

1. Naming of `StaticBindingJsNative` feels very specific, and doesn't match the intent necessarily.
2. We're adding files and putting outdated Lightbend copyright data at the top, just to satisfy CI. We should do something different in the future.

BillyAutrey:
@mkurz I'll hold off on merging, wanted to give you another day to come back and make sure I didn't miss anything. I'll merge tomorrow if you have no other thoughts.

mkurz:
will take a look tonight

lolgab:
@BillyAutrey I can rename `StaticBindingJsNative` to `StaticBindingNonJvm` right now.

BillyAutrey:
@lolgab up to you! Like I said I feel it's a nit, and an easy one to fix later if we want to discuss it further.

lolgab:
> @lolgab up to you! Like I said I feel it's a nit, and an easy one to fix later if we want to discuss it further.

Since we are waiting for tomorrow either way, it doesn't cost me anything to update it :)

lolgab:
Thank you very much for reviewing and approving the PR!

mkurz (approved):
LGTM, rebased and [upgraded versions](https://repo1.maven.org/maven2/org/scala-native/nscplugin_3.3.0-RC5/), since we are now on Scala 3.3.0-RC5.

play-json/native/src/main/scala/StaticBinding.scala
mkurz: I will do that in a upcoming pr.

Issue #107, opened 2017-09-20
Double values with 0 fraction written without fraction


francisdb:
play-json 2.6.5

**issue**

Play json will serialize `JsNumber(BigDecimal("1.0"))` to `1` where this should be `1.0`

**reproducer**

```scala
  val mapper = new ObjectMapper()
  val jacksonObject = mapper.createObjectNode()
  jacksonObject.put("test", 1d)
  println(mapper.writerWithDefaultPrettyPrinter.writeValueAsString(jacksonObject))
  
  val playObject2 = obj(
    "test" -> 1d
  )
  println(prettyPrint(playObject2))
```

output looks like this
```js
// correct with jackson
{
  "test" : 1.0
}
// incorrect with play
{
  "test" : 1
}
```

The code explicitly strips trailing zeros, this is a loss of precision:
https://github.com/playframework/play-json/blob/master/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala#L66

**usecase** 

Writing to KairosDB will create a schema for Real values if the incoming datapoint contains a `.`, otherwise data is seen as Integer.

gmethvin:
This sounds more like a feature request than a bug, and I don't know if your use case is the right approach. JSON itself does not distinguish between integers and real numbers. There is only one number type.

It is common for JSON implementations to normalize the representation of numbers. I can see the value of pretty-printing JSON for presentational purposes, and I think it'd be reasonable for play-json to make that configurable.

francisdb:
KairosDB (time series database) wants to see json values with `.0` to make a difference between long and double columns. And as you can see above the underlying library play-json is using (jackson) supports this.

[RFC 7159](https://tools.ietf.org/html/rfc7159#section-6) does not mention anything about normalization. I have a hard time to believe this would be common unless in the language itself you can not make a difference (js).

I see no reason why play-json could not keep the precision and represent the numbers as printed by BigDecimal or with at least one fractional digit if the source of the number was a double. This will not break anything for existing users. The only reason that comes to mind would be a (marginal) reduction of bandwidth which the user could mitigate by replacing doubles without fraction by integers

```scala
scala> val x = BigDecimal(1d)
x: scala.math.BigDecimal = 1.0

scala> val x = BigDecimal(1)
x: scala.math.BigDecimal = 1

scala> val x = BigDecimal(1.00d)
x: scala.math.BigDecimal = 1.0

scala> val x = BigDecimal("1.00")
[...]

PR #831, opened 2022-11-22
Add option to preserve zero decimals

trbogart:
# Pull Request Checklist

* [X] Have you added copyright headers to new files?
 - Updated documentation in code

## Fixes

Fixes [#107](https://github.com/playframework/play-json/issues/107) by adding a configuration option to preserve zero decimal numbers from being rounded to a whole number (for example, 1.0 will not be converted to 1). 

In particular, we want to use this to prevent ElasticSearch from creating a long dynamic mapping for a field that should be a double. It's not really a bug (as described in that issue), so the old behavior is preserved by default. 

My first approach was to serialize this in our code base. That works, but isn't ideal. It required copying a fair bit of play-json code to override just a single line of code. Also, our code serializes Json in dozens of places, all of which had to be changed to use the new methods.

## Purpose

This adds an option to preserve a .0 for decimal numbers that are rounded to a whole number (e.g. 1.0 to 1). This distinction is important in some situations, such as ES dynamic mapping or KairosDB schema creation.

## Background Context

It might be harmless to do the new behavior by default, but the original issue did mention a preference for opting in to this behavior, which is reasonable.

## References

https://github.com/playframework/play-json/issues/107


play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -257,7 +267,15 @@ private[jackson] class PlaySerializers(parserSettings: JsonParserSettings) exten
 }
 
 private[json] object JacksonJson {
-  private lazy val mapper = (new ObjectMapper).registerModule(new PlayJsonModule(JsonParserSettings.settings))
+  /**
+   * Instance used to serialize and deserialize JSON. This is configured with system properties, but can be
+   * overridden for testing.
+   */
+  var instance: JacksonJson = JacksonJson(JsonParserSettings.settings)
```
trbogart, line 274: Global static state isn't ideal, but this isn't globally exposed and adds a hook for testing.

I think it would be better to use implicit state, but I couldn't figure out a way to do that without breaking backwards compatibility.

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -69,6 +69,16 @@ private[jackson] class JsValueSerializer(parserSettings: JsonParserSettings) ext
   import com.fasterxml.jackson.databind.node.BigIntegerNode
   import com.fasterxml.jackson.databind.node.DecimalNode
 
+  private def stripTrailingZeros(bigDec: JBigDec): JBigDec = {
```
trbogart, line 72: This is the core change.

trbogart, line 72: This is the core change.

gmethvin:
@cchantep Thoughts? The general approach seems reasonable to me.

play-json/jvm/src/main/scala/play/api/libs/json/JsonParserSettings.scala
```diff
@@ -18,15 +23,21 @@ import scala.util.control.NonFatal
  */
 final case class BigDecimalParseSettings(
     mathContext: MathContext = MathContext.DECIMAL128,
-    scaleLimit: Int,
-    digitsLimit: Int
+    scaleLimit: Int = defaultScaleLimit,
```
cchantep, line 26: Changing defaults is probably a breaking change ... See mima

cchantep (changes requested):

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -69,6 +69,16 @@ private[jackson] class JsValueSerializer(parserSettings: JsonParserSettings) ext
...
+    val stripped = bigDec.stripTrailingZeros
+    if (parserSettings.bigDecimalSerializerSettings.preserveZeroDecimal && bigDec.scale > 0 && stripped.scale == 0) {
+      // restore .0 if rounded to a whole number
+      stripped.setScale(1)
```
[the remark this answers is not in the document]
trbogart: I don't see a straightforward way to do that with the BigDecimal API.

It looks like `scala.BigDecimal` has an `isWhole` method, but it actually calls `stripTrailingZeros` anyway (at least for the cases where this would be triggered), so it wouldn't help.

Fortunately, increasing the scale is a fairly straightforward operation.

trbogart:
I'm not sure why mima is still failing.  I'll have to come back to that next week.

gmethvin (commented):
MiMa is failing because you've changed the case class constructor to have 3 arguments rather than 2, which also affects the apply and copy methods generated by the case class. 

Arguably it was a bad idea to make `JsonParserSettings` a case class in #191, as it makes it inconvenient to add new settings to the case class without breaking compatibility. We can still technically work around this by manually adding `copy`, `apply` and `this` methods with the old arguments, but it ends up being pretty ugly, and we will have to do this again every time we add another field to the case class.

So a few options are:
 1. Add in the extra methods for binary compatibility—these could be made `private[json]` so they are only visible publicly in the binary representation.
 2. Option 1, plus we make both the constructor and copy method private, so in the future when adding config we only need to worry about the `apply` method.
 3. Switch to using a (preferably sealed) trait for configuration, deprecate the old case class config, and create a new way to construct instance of the trait for config.
 4. Wait to introduce this breaking change in the next "major" version of play-json—which I think would be 2.10.0, and is coming up pretty soon.

Option 2 seems like the best balance between amount of work now and extensibility in the future.

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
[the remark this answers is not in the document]
cchantep: Seems quitd weird to me

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -69,6 +70,27 @@ class JsonSpec extends org.specs2.mutable.Specification {
...
+    )
+    JacksonJson(JsonParserSettings(defaultParserSettings, serializerSettings))
+  }
+
+  def withJacksonJson[T](jacksonJson: JacksonJson)(f: () => T) = {
+    val oldInstance = JacksonJson.set(jacksonJson)
+    try {
+      JacksonJson.get = jacksonJson
+      f.apply()
+    } finally {
```
cchantep, line 89: Silent error is generally not a good idea, even in tests

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -6,30 +6,28 @@ package play.api.libs.json.jackson
 
 import java.io.InputStream
 import java.io.StringWriter
-
```
cchantep, line 9: Please keep formating

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -257,7 +265,22 @@ private[jackson] class PlaySerializers(parserSettings: JsonParserSettings) exten
...
-  private lazy val mapper = (new ObjectMapper).registerModule(new PlayJsonModule(JsonParserSettings.settings))
+  val defaultInstance: JacksonJson = JacksonJson(JsonParserSettings.settings)
+
+  private val ref: AtomicReference[JacksonJson] = new AtomicReference[JacksonJson](defaultInstance)
+
+  /**
+   * Instance used to serialize and deserialize JSON. This is configured with system properties, but can be
+   * overridden for testing.
+   */
+  var get: JacksonJson = ref.get
```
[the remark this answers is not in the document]
trbogart: Oops

play-json/jvm/src/main/scala/play/api/libs/json/JsonConfig.scala
```diff
@@ -0,0 +1,207 @@
+/*
+ * Copyright (C) 2009-2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package play.api.libs.json
```
trbogart, line 5: Renamed from `JsonParserSettings`.

I ended up using sealed traits, which I called `JsonConfig`, `BigDecimalParseConfig`, and `BigDecimalSerializerConfig`.

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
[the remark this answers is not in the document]
trbogart: I don't see a reasonable way to do this with `BigDecimal` itself.

I tried multiplying by 10, then stripping zeros, then dividing by 10, but `stripTrailingZeros` will actually strip all trailing zeros, even before the decimal point.

Another option is to strip the trailing zeros from the string instead of the `BigDecimal`. I'm not opposed to that, but it seems clunkier.

Given all that, I think setting the scale is the best option.

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
trbogart: Used `AtomicReference`

gmethvin:
@trbogart thanks for the configuration changes. @cchantep what do you think about this approach?

trbogart:
@gmethvin and @cchantep, this is ready for review again.

play-json/jvm/src/main/scala/play/api/libs/json/JsonConfig.scala
```diff
@@ -0,0 +1,210 @@
...
+
+  private[json] def parseMathContext(key: String): MathContext = sys.props.get(key).map(_.toLowerCase) match {
+    case Some("decimal128") => MathContext.DECIMAL128
+    case Some("decimal64")  => MathContext.DECIMAL64
+    case Some("decimal32")  => MathContext.DECIMAL32
+    case Some("unlimited")  => MathContext.UNLIMITED
+    case _                  => defaultMathContext
+  }
+
+  private[json] def parseNum[T](key: String, default: T)(f: String => T): T =
```
cchantep, line 140:
```suggestion
  private[json] def numericProp[T](key: String, default: T)(f: String => T): T =
```

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -257,7 +269,22 @@ private[jackson] class PlaySerializers(parserSettings: JsonParserSettings) exten
 }
 
 private[json] object JacksonJson {
-  private lazy val mapper = (new ObjectMapper).registerModule(new PlayJsonModule(JsonParserSettings.settings))
+  val defaultInstance: JacksonJson = JacksonJson(JsonConfig.settings)
+
+  private val ref: AtomicReference[JacksonJson] = new AtomicReference[JacksonJson](defaultInstance)
```
cchantep, line 274: So it add overhead for all runtime whereas it's only useful for testing?
Couldn't the internal set logic and its synchronization be only in tests?

gmethvin: I agree if this is only for tests it probably doesn't need to be atomic. This is an internal API so we can control how it's used.

Also wondering if we could keep `JacksonJson` an object and just have a `setConfig` in there?

trbogart: Having the case class also means that any given operations will use a consistent set of mapper and factory, so there's not really a need for synchronization. Running tests concurrently could still give unexpected results (since it relies on static state), but the atomic reference didn't actually help with that anyway.

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
trbogart: Reverted the `AtomicReference` change, as mentioned in another thread.

All serialization or parsing operations will be done on a single immutable `JacksonJson` object, and any changes here will not affect any such operations in flight. Concurrent tests that change the config could produce unexpected results, but the only way to avoid that is to refactor the library to avoid static state. I don't think that's a problem in practice, though.

gmethvin (approved):

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
cchantep: Not resolved

trbogart: Fixed

gmethvin:
@cchantep do you have any other concerns or are we good to merge this?

trbogart:
Actually, there is another thing before merging. I added some deprecated annotations, and I wasn't sure what to use for the "since" field. I used "2.10.0", but I'm not sure if that is correct.

Also, what is the timeline for getting this released?

Thanks!

@gmethvin and @cchantep

mkurz:
> Also, what is the timeline for getting this released?

I didn't follow this conversations, however if  @gmethvin and @cchantep merge this I can take over and cut a new release afterwards.

gmethvin:
I think we want to backport this to 2.9.x, so deprecating in `2.9.4` would make sense? I believe play-json 2.9.x is usable with Play 2.8?

mkurz:
> I think we want to backport this to 2.9.x, so deprecating in `2.9.4` would make sense? I believe play-json 2.9.x is usable with Play 2.8?

Right now Play 2.8 [ships with play-json 2.8.x](https://github.com/playframework/playframework/blob/f71c8c9d482b2d75f9edc01b79bf896cb49ade6e/project/Dependencies.scala#L15), however I think it should be possible to manually upgrade to play-json 2.9.x, I think some people did that already, not sure though.

cchantep (approved):

trbogart:
2.8.x backport PR: https://github.com/playframework/play-json/pull/840
2.9.x backport PR: https://github.com/playframework/play-json/pull/839

PR #843, opened 2023-01-27
Allow setting configuration programmatically

trbogart:
# Pull Request Checklist

## Fixes

Fixes #xxxx

## Purpose

Extends https://github.com/playframework/play-json/pull/831 to allow setting configuration programmatically.

## Background Context

Why did you take this approach?

## References

### DIFF
--- a/play-json/jvm/src/main/scala/play/api/libs/json/JsonConfig.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/JsonConfig.scala
@@ -194,8 +194,9 @@ object JsonConfig {
   private[json] def loadPreserveZeroDecimal: Boolean =
     prop(preserveZeroDecimalProperty, defaultPreserveZeroDecimal)(_.toBoolean)
 
-  // Default settings, which can be controlled with system properties.
-  // Can be configured with system properties or programmatically with setConfig()
+  /**
+   * Default settings, which can be controlled with system properties or set programmatically with [[setConfig]]
+   */
   val settings: JsonConfig =
     JsonConfig(
       BigDecimalParseConfig(loadMathContext, loadScaleLimit, loadDigitsLimit),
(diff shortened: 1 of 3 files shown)


play-json/jvm/src/main/scala/play/api/libs/json/JsonConfig.scala
```diff
@@ -186,9 +193,13 @@ object JsonConfig {
...
     prop(preserveZeroDecimalProperty, defaultPreserveZeroDecimal)(_.toBoolean)
 
-  // Default settings, which can be controlled with system properties.
-  // To override, call JacksonJson.setConfig()
-  val settings: JsonConfig =
+  /** Override configuration */
+  def setConfig(jsonConfig: JsonConfig): Unit = JacksonJson.initConfig(jsonConfig)
+
+  /** Current configuration. */
+  def settings: JsonConfig = JacksonJson.get.jsonConfig
```
trbogart, line 200: This should probably be called `get` or `getConfig`, but I'm leaving it for backwards compatibility.

play-json/jvm/src/main/scala/play/api/libs/json/JsonConfig.scala
```diff
@@ -186,9 +193,13 @@ object JsonConfig {
   private[json] def loadPreserveZeroDecimal: Boolean =
     prop(preserveZeroDecimalProperty, defaultPreserveZeroDecimal)(_.toBoolean)
 
-  // Default settings, which can be controlled with system properties.
-  // To override, call JacksonJson.setConfig()
-  val settings: JsonConfig =
+  /** Override configuration */
+  def setConfig(jsonConfig: JsonConfig): Unit = JacksonJson.initConfig(jsonConfig)
```
trbogart, line 197: There was concern in the original PR about thread-safety for exposing this publicly. However, I don't think that's really an issue. This config is only used to initialize `JacksonJson`, so any given operation will use a consistent set of classes.

cchantep: I don't think exposing such mutation in the API is a good idea

cchantep (changes requested):

PR #987, opened 2024-02-25
Close #700 - Safe Json.parse alternatives

### LINKED ISSUE #700: Add a JSON parsing API that returns Try or similar type instead of throwing exception
Currently, when `JSON.parse` processes invalid JSON, it will throw an exception. This is not documented and can be misleading for some users that will forget to catch the exceptions.

It would be ideal to have a new API (not to break compatibility) with a signature like 

```scala

def tryParse(input: String): Try[JsValue] // or with Either or similar type that encodes error

def tryParse(input: InputStream): Try JsValue]

def tryParse(input: Array[Byte]): Try[JsValue]
```
in order to use the type system to prevent user mistakes.

If it is not possible at all to add a new API, at least it would be important to document that exceptions can be thrown.

### Play JSON Version
2.9.x

### API
Scala

cchantep:
# Pull Request Checklist

## Fixes

Fixes #700

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/Json.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/Json.scala
@@ -4,33 +4,46 @@
 
 package play.api.libs.json
 
+import java.io.InputStream
+
 import scala.collection.mutable.{ Builder => MBuilder }
 
-import java.io.InputStream
+import scala.util.Try
 
 /**
  * @define jsonParam @param json the JsValue to convert
- * @define returnStringRepr A String with the json representation
+ * @define parseDescription Parses the input JSON, and returns it as a [[JsValue]]
+ * @define stringRepr String with the json representation
  */
 sealed trait JsonFacade {
 
   /**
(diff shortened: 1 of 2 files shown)


mkurz:
@cchantep Binary Compatibility checks fail, so probably this has to wait for play-json 3.1.0 (?)

build.sbt
```diff
@@ -45,8 +45,22 @@ val previousVersion: Option[String] = Some("3.0.0")
 // Do not check for previous JS artifacts for upgrade to Scala.js 1.0 because no sjs1 artifacts exist
 def playJsonMimaSettings = Seq(
   mimaPreviousArtifacts := previousVersion.map(organization.value %%% moduleName.value % _).toSet,
-  mimaBinaryIssueFilters ++= Seq(
-  ),
+  mimaBinaryIssueFilters ++= {
+    val missingMethodInOld: ProblemFilter = {
```
cchantep, line 49: Adding new feature doesn't break backward compat

PR #993, opened 2024-03-03
Fix #974 – Scala 3 macro cannot find Writes for Seq[Map[String, T]]

### LINKED ISSUE #974: Scala 3 macro cannot find Writes for Seq[Map[String, T]]
## Code

```scala
import play.api.libs.json._

case class Result(status: String)
object Result {
  implicit val format: OFormat[Result] = Json.format[Result]
}

case class Response(results: Seq[Map[String, Result]])
object Response {
  implicit val format: OFormat[Response] = Json.format[Response]
}

val response = Response(results = Seq(Map("123" -> Result("ok"))))
println(Json.toJson(response))
```

## Scala 2.13.12 (works)

```
-> % scala-cli -S 2.13.12 playJsonSeqMapMacro.sc
Compiling project (Scala 2.13.12, JVM (17))
Compiled project (Scala 2.13.12, JVM (17))
{"results":[{"123":{"status":"ok"}}]}
```

## Scala 3.3.1 (fails)

```
-> % scala-cli -S 3.3.1 playJsonSeqMapMacro.sc
Compiling project (Scala 3.3.1, JVM (17))
[error] ./playJsonSeqMapMacro.sc:12:44
[error] Instance not found: play.api.libs.json.Writes[scala.collection.immutable.Seq[scala.collection.immutable.Map[String, playJsonSeqMapMacro_.Result]]]
[error]   implicit val format: OFormat[Response] = Json.format[Response]
[error]                                            ^^^^^^^^^^^^^^^^^^^^^
Error compiling project (Scala 3.3.1, JVM (17))
Compilation failed
```

sgodbillon:
Moved the deprecated method `DefaultWrites.mapWrites` into `LowPriorityWrites` to get rid of the ambiguity.

# Pull Request Checklist

## Fixes

Fixes #974 – Scala 3 macro cannot find Writes for Seq[Map[String, T]]

## Purpose

What does this PR do?

## Background Context

Why did you take this approach?

## References

### DIFF
--- a/play-json/shared/src/test/scala/play/api/libs/json/WritesSharedSpec.scala
+++ b/play-json/shared/src/test/scala/play/api/libs/json/WritesSharedSpec.scala
@@ -84,6 +84,10 @@ final class WritesSharedSpec extends AnyWordSpec with Matchers {
     "write lazy maps" in {
       Json.toJson(Map("a" -> 1).map(kv => kv._1 -> (kv._2 + 1))).mustEqual(Json.obj("a" -> 2))
     }
+
+    "write a map nested in a seq" in {
+      Json.toJson(Seq(Map("a" -> 1))).mustEqual(Json.arr(Json.obj("a" -> 1)))
+    }
   }
 
   "Iterable writes" should {
--- a/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
@@ -366,7 +366,7 @@ trait DefaultWrites extends LowPriorityWrites with EnumerationWrites {
    * Serializer for Map[String,V] types.
    */
   @deprecated("Use `genericMapWrites`", "2.8.0")
-  implicit def mapWrites[V: Writes]: OWrites[MapWrites.Map[String, V]] = MapWrites.mapWrites
+  def mapWrites[V: Writes]: OWrites[MapWrites.Map[String, V]] = MapWrites.mapWrites
 
   implicit def keyMapWrites[K: KeyWrites, V: Writes, M[K, V] <: MapWrites.Map[K, V]]: OWrites[M[K, V]] = {
     val kw = implicitly[KeyWrites[K]]


sgodbillon:
About binary compatibility – there is another approach we can take that would be binary compatible: keep `mapWrites` where it is but drop the `implicit` keyword. That breaks source compatibility, but only when it is imported explicitly so the impact would be minimal. Let me know what you think!

play-json/shared/src/test/scala/play/api/libs/json/WritesSharedSpec.scala
```diff
@@ -84,6 +84,9 @@ final class WritesSharedSpec extends AnyWordSpec with Matchers {
     "write lazy maps" in {
       Json.toJson(Map("a" -> 1).map(kv => kv._1 -> (kv._2 + 1))).mustEqual(Json.obj("a" -> 2))
     }
+    "write a map nested in a seq" in {
```
cchantep, line 87:
```suggestion

    "write a map nested in a seq" in {
```

cchantep:
Seems it needs to be updated about the MiMa check: https://github.com/playframework/play-json/actions/runs/8133336570/job/22224801331?pr=993#step:6:226

play-json/shared/src/test/scala/play/api/libs/json/WritesSharedSpec.scala
sgodbillon: done

sgodbillon:
> Seems it needs to be updated about the MiMa check: https://github.com/playframework/play-json/actions/runs/8133336570/job/22224801331?pr=993#step:6:226

@cchantep Please see my comment above :)

> About binary compatibility – there is another approach we can take that would be binary compatible: keep mapWrites where it is but drop the implicit keyword. That breaks source compatibility, but only when it is imported explicitly so the impact would be minimal. Let me know what you think!

cchantep:
> > Seems it needs to be updated about the MiMa check: https://github.com/playframework/play-json/actions/runs/8133336570/job/22224801331?pr=993#step:6:226
> 
> @cchantep Please see my comment above :)
> 
> > About binary compatibility – there is another approach we can take that would be binary compatible: keep mapWrites where it is but drop the implicit keyword. That breaks source compatibility, but only when it is imported explicitly so the impact would be minimal. Let me know what you think!

As soon as we speak of 2.8, without backport, I'm fine with either excluding the MiMa warn in build or keeping the function without `implicit`.

sgodbillon:
> As soon as we speak of 2.8, without backport, I'm fine with either excluding the MiMa warn in build or keeping the function without implicit.

Yes, that's only for 2.10. I opted for removing the implicit keyword

cchantep (approved):

mkurz:
> > As soon as we speak of 2.8, without backport, I'm fine with either excluding the MiMa warn in build or keeping the function without implicit.
> 
> Yes, that's only for 2.10. I opted for removing the implicit keyword

Sorry I don't follow, should this fix backported to 2.10.x and be part of a 2.10.5 release?

mkurz:
OK, if I understood correctly you want that be part of 2.10.5, so let's backport.

mkurz:
@Mergifyio backport 2.10.x

PR #1020, opened 2024-04-29
Revert "sbt-scalajs, scalajs-compiler, ... 1.16.0 (was 1.15.0)"

mkurz:
Reverts playframework/play-json#1006

We did not merge this in twirl, so let's do the same for play-json, see comments starting here: https://github.com/playframework/twirl/pull/750#issuecomment-2076890907

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -146,7 +146,7 @@ lazy val `play-json` = crossProject(JVMPlatform, JSPlatform, NativePlatform)
       libraryDependencies ++= Seq(
         "org.scalatest"     %%% "scalatest"       % "3.2.18"   % Test,
         "org.scalatestplus" %%% "scalacheck-1-16" % "3.2.14.0" % Test,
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -146,7 +146,7 @@ lazy val `play-json` = crossProject(JVMPlatform, JSPlatform, NativePlatform)
       libraryDependencies ++= Seq(
         "org.scalatest"     %%% "scalatest"       % "3.2.18"   % Test,
         "org.scalatestplus" %%% "scalacheck-1-16" % "3.2.14.0" % Test,
-        "org.scalacheck"    %%% "scalacheck"      % "1.17.1"   % Test,
+        "org.scalacheck"    %%% "scalacheck"      % "1.17.0"   % Test,
```
mkurz, line 149: Need to downgrade because scalacheck patch release upgraded to sbt-scalajs 1.16 which is incompatible: https://github.com/typelevel/scalacheck/commit/8caf0b1de4bce75feefdc39c18886367c8530a03

PR #968, opened 2024-01-16
Run tests on Java 21

mkurz:
Should fail according to
- #929

PR #929, opened 2023-10-09
fix JsonMemoryFootprintSpec with JDK 21

xuwei-k:
# Pull Request Checklist

* [ ] Have you read through the [contributor guidelines](https://www.playframework.com/contributing)?
* [ ] Have you [squashed your commits](https://www.playframework.com/documentation/latest/WorkingWithGit#Squashing-commits)?
* [ ] Have you added copyright headers to new files?
* [ ] Have you updated the documentation?
* [ ] Have you added tests for any changed functionality?

## Purpose

pending some test if JDK 21 because fail.

## Background Context

https://github.com/openjdk/jdk/commit/17ce0976e442d5fabb14daed40fa9a768989f02e#diff-cdf6d0587f1021c14e1af0d9cab581de2809cd1dbcf8bd87f5a5a304dc087263R333

add `int putMode` since JDK 21.

inspect `java.util.LinkedHashMap` by jol

### JDK 17 

```
java.util.LinkedHashMap object internals:
 OFFSET  SIZE       TYPE DESCRIPTION                    VALUE
      0     4            (object header)                01 00 00 00 (00000001 00000000 00000000 00000000) (1)
      4     4            (object header)                00 00 00 00 (00000000 00000000 00000000 00000000) (0)
      8     4            (object header)                d8 65 11 00 (11011000 01100101 00010001 00000000) (1140184)
     12     4        Set AbstractMap.keySet             (access denied)
     16     4 Collection AbstractMap.values             (access denied)
     20     4        int HashMap.size                   (access denied)
     24     4        int HashMap.modCount               (access denied)
     28     4        int HashMap.threshold              (access denied)
     32     4      float HashMap.loadFactor             (access denied)
     36     4     Node[] HashMap.table                  (access denied)
     40     4        Set HashMap.entrySet               (access denied)
     44     1    boolean LinkedHashMap.accessOrder      (access denied)
     45     3            (alignment/padding gap)        N/A
     48     4      Entry LinkedHashMap.head             (access denied)
     52     4      Entry LinkedHashMap.tail             (access denied)
Instance size: 56 bytes
Space losses: 3 bytes internal + 0 bytes external = 3 bytes total
```

### JDK 21

```
java.util.LinkedHashMap object internals:
 OFFSET  SIZE       TYPE DESCRIPTION                    VALUE
      0     4            (object header)                01 00 00 00 (00000001 00000000 00000000 00000000) (1)
[...]

PR #1033, opened 2024-05-03
Run tests with Java 21 also

mkurz:
I think back then it was not available in the coursier index.


mkurz:
We have
- #929
- #968

already...

Issue #1038, opened 2024-05-09
JSON field ordering bug with Scala 3


levinson:
### Play JSON Version (2.5.x / etc)

Affected versions:
* `"com.typesafe.play" %% "play-json" % "2.10.5"`
* `"org.playframework" %% "play-json" % "3.0.3"`

### API (Scala / Java / Neither / Both)

Scala

### Operating System (Ubuntu 15.10 / MacOS 10.10 / Windows 10)

MacOS 14.3.1

### JDK (Oracle 1.8.0_72, OpenJDK 1.8.x, Azul Zing)

java version "17.0.10" 2024-01-16 LTS
Java(TM) SE Runtime Environment (build 17.0.10+11-LTS-240)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.10+11-LTS-240, mixed mode, sharing)

### Expected Behavior

Expected JSON ordering to be preserved when serializing model class.

### Actual Behavior

JSON field ordering is not preserved when compiling using Scala 3

### Reproducible Test Case

I created the following project to demonstrate the issue: https://github.com/levinson/play-json-scala3-ordering-bug

Posting the source code here as well for reference:

build.sbt:
```sbt
scalaVersion := "2.13.14"
// TODO: Test fails when using Scala 3
//scalaVersion := "3.3.3"

libraryDependencies ++= Seq(
  "com.typesafe.play" %% "play-json" % "2.10.5",
  //"org.playframework" %% "play-json" % "3.0.3", // same behavior
  "org.scalatest" %% "scalatest" % "3.2.18" % Test
)
```
JsonSerializersTest.scala
```scala
import org.scalatest.matchers.should.Matchers
import org.scalatest.wordspec.AnyWordSpec
import play.api.libs.json._

case class Model(metadata: Option[String], region: String)

object JsonSerializers {
  implicit val modelFmt: Format[Model] = Json.format[Model]
}

class JsonSerializersTest extends AnyWordSpec with Matchers {

  import JsonSerializers._

  "JsonSerializers" should {
    "preserve ordering" in {
      val model = Model(Some("metadata"), "region")
      val serialized = Json.stringify(Json.toJson(model))
      println("serialized is " + serialized)
      serialized shouldBe """{"metadata":"metadata","region":"region"}"""
    }
  }
}
```

gbarmashiahflir:
Happens here as well.
Having for example this class:

```
object MetricStatistics{
[...]

PR #1042, opened 2024-05-11
Add test for field ordering

levinson:
# Pull Request Checklist

* [X] Have you added copyright headers to new files? N/A
* [X] Have you updated the documentation? N/A
* [X] Have you added tests for any changed functionality? N/A

## Purpose

This PR demonstrates #1038. The new test fails due to serialized field order when compiled with Scala 3:
```
[info] - should preserve field ordering *** FAILED *** (33 milliseconds)
[info]   "{"[a":42,"m1":"foo","b":"baz","c":"bar","z":1.2]}" did not equal "{"[m1":"foo","c":"bar","z":1.2,"a":42,"b":"baz"]}" (JsonSharedSpec.scala:454)
```


play-json/shared/src/test/scala/play/api/libs/json/JsonSharedSpec.scala
```diff
@@ -443,6 +447,12 @@ class JsonSharedSpec extends AnyWordSpec with Matchers with org.scalatestplus.sc
 
       test.toString.mustEqual(js.parse(req).toString) // ).ignoreSpace
     }
+
+    "preserve field ordering" in {
+      val model      = Model(Some("foo"), "bar", Some(1.2), 42, Some("baz"))
+      val serialized = Json.stringify(Json.toJson(model))
+      serialized.mustEqual("""{"m1":"foo","c":"bar","z":1.2,"a":42,"b":"baz"}""")
```
cchantep, line 454:
```suggestion

      serialized.mustEqual("""{"m1":"foo","c":"bar","z":1.2,"a":42,"b":"baz"}""")
```

jonaskoelker-jypo:
Would it make sense to test the field ordering at the `JsObject` level in addition to the `String` level?

I note the comment on the `fields` method on `JsObject`: it gives the fields _in order_. Based on this, I would—naively?—expect the `Seq` returned by `fields` to be in declaration order when using the `Json.writes` macro and friends. It would be a real nicety to have a test for that as well.

https://github.com/playframework/play-json/blob/85a8aa03a506e8dbf57cb8f6afcf580a89e6fdfd/play-json/shared/src/main/scala/play/api/libs/json/JsValue.scala#L116-L123

[On the other hand: I can argue that a `JsObject` doesn't contain any information about the field declaration order except the ordering of the `ImmutableLinkedHashMap` in `underlying`. Ergo: if it's wrong at the `JsObject` level there's no way `Json.stringify` can correct it, and thus this test case implicitly also verifies that `underlying` is ordered correctly. That feels a bit indirect for my tastes—I lean towards an independent test of that—but I figured it deserved to be mentioned.]

mkurz:
@Mergifyio rebase

mkurz:
https://github.com/Mergifyio backport 3.0.x 2.10.x

PR #1047, opened 2024-06-05
Make format generation idempotent

reimai:
## Fixes

Code generated by JsMacroImpl is slightly different between compilations, the order of sealed trait's subclasses keeps changing, because they are collected into a Set in ```directKnownSubclasses``` method. I've simply changed that to a LinkedHashSet.
We're using bazel as a build tool and it's important for build to produce an exact same file for the same input, otherwise it's dependent targets are re-evaluated.

## Purpose

Have better caching in bazel project using play framework.


mkurz (approved):
Seems legit.
Will be part of Play's next patch release.

mkurz:
@Mergifyio backport 2.10.x

PR #1074, opened 2024-09-05
avoid reparsing numbers when serializing

pjfanning:
Alternative to #1073 

This change avoids creating NumericNodes and parsing strings to numbers which ultimately have to be turned back to strings to write them.

I've had to leave the BigIntegerNode usage in there. Removing it breaks some tests. I've debugged and it is down to how writeNumber(String) is implemented in jackson-databind TokenBuffer. We would need to change this class in jackson-databind to fix this (and that is not likely to happen as it will probably cause other issues). Basically, writeNumber(String) is assumed to be a non-integer in TokenBuffer.

### DIFF
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -100,8 +100,8 @@ private[jackson] class JsValueSerializer(jsonConfig: JsonConfig) extends JsonSer
           json.writeNumber(raw)
         else
           json match {
-            case _: TokenBuffer =>
-              json.writeTree(new BigIntegerNode(new BigInteger(raw)))
+            case tb: TokenBuffer =>
+              tb.writeNumber(raw, true)
             case _ =>
               json.writeNumber(raw)
           }
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -6,7 +6,6 @@ package play.api.libs.json.jackson
 
 import java.io.InputStream
 import java.io.StringWriter
-import java.math.BigInteger
 
 import scala.annotation.switch
 import scala.annotation.tailrec
@@ -28,7 +27,6 @@ import com.fasterxml.jackson.databind.`type`.TypeFactory
 import com.fasterxml.jackson.databind.deser.Deserializers
 import com.fasterxml.jackson.databind.json.JsonMapper
 import com.fasterxml.jackson.databind.module.SimpleModule
-import com.fasterxml.jackson.databind.node.BigIntegerNode
 import com.fasterxml.jackson.databind.ser.Serializers
 import com.fasterxml.jackson.databind.util.TokenBuffer
 
--- a/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
@@ -6,6 +6,7 @@ package play.api.libs.json.jackson
 
 import java.io.InputStream
 import java.io.StringWriter
+import java.math.BigInteger
 
 import scala.annotation.switch
 import scala.annotation.tailrec
@@ -27,7 +28,9 @@ import com.fasterxml.jackson.databind.`type`.TypeFactory
 import com.fasterxml.jackson.databind.deser.Deserializers
 import com.fasterxml.jackson.databind.json.JsonMapper
 import com.fasterxml.jackson.databind.module.SimpleModule
+import com.fasterxml.jackson.databind.node.BigIntegerNode
 import com.fasterxml.jackson.databind.ser.Serializers
+import com.fasterxml.jackson.databind.util.TokenBuffer
 
 import play.api.libs.json._
 
@@ -68,11 +71,7 @@ sealed class PlayJsonMapperModule(jsonConfig: JsonConfig) extends SimpleModule("
 // -- Serializers.
 
 private[jackson] class JsValueSerializer(jsonConfig: JsonConfig) extends JsonSerializer[JsValue] {
-  import java.math.BigInteger
   import java.math.{ BigDecimal => JBigDec }
-
-  import com.fasterxml.jackson.databind.node.BigIntegerNode
-  import com.fasterxml.jackson.databind.node.DecimalNode
 
   private def stripTrailingZeros(bigDec: JBigDec): JBigDec = {
     val stripped = bigDec.stripTrailingZeros
@@ -97,10 +96,15 @@ private[jackson] class JsValueSerializer(jsonConfig: JsonConfig) extends JsonSer
         val stripped = stripTrailingZeros(v.bigDecimal)
         val raw      = if (shouldWritePlain) stripped.toPlainString else stripped.toString
 
-        if (raw.indexOf('E') < 0 && raw.indexOf('.') < 0)
-          json.writeTree(new BigIntegerNode(new BigInteger(raw)))
+        if (raw.exists(c => c == 'E' || c == '.'))
+          json.writeNumber(raw)
         else
-          json.writeTree(new DecimalNode(new JBigDec(raw)))
+          json match {
+            case _: TokenBuffer =>
+              json.writeTree(new BigIntegerNode(new BigInteger(raw)))
+            case _ =>
+              json.writeNumber(raw)
+          }
       }
 
       case JsString(v)  => json.writeString(v)


play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -125,6 +123,16 @@ private[jackson] class JsValueSerializer(jsonConfig: JsonConfig) extends JsonSer
       case JsNull => json.writeNull()
     }
   }
+
+  private def containsEOrDot(s: String): Boolean = {
```
cchantep, line 127: `s.exists(...)`

play-json/jvm/src/main/scala/play/api/libs/json/jackson/StringBasedNumericNode.scala
```diff
@@ -0,0 +1,46 @@
...
+  override def asText(): String = text
+
+  override def asToken(): JsonToken = JsonToken.VALUE_NUMBER_FLOAT
+
+  override def isIntegralNumber: Boolean = false
+
+  override def serialize(jgen: JsonGenerator, ctxt: SerializerProvider): Unit =
+    jgen.writeNumber(text)
+
+  override def numberValue: Number = throw new NotImplementedError
```
cchantep, line 31: So doesn't look like a `NumericNode` for now

cchantep (changes requested):

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
pjfanning: Are you suggesting a rename for the the 'containsEOrDot' function or a change in its implementation?
I am not aware of a method on String called 'exists'. I checked the Javadocs.

play-json/jvm/src/main/scala/play/api/libs/json/jackson/StringBasedNumericNode.scala
pjfanning: I describe in the PR description as to why I don't need this method to be implemented. I only really need the serialize method to be implemented. This is not a general purpose public class - it is an internal class written this way to avoid extra conversions.

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
pjfanning: actually, I see your point now - I will make the change you requested

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -125,6 +129,9 @@ private[jackson] class JsValueSerializer(jsonConfig: JsonConfig) extends JsonSer
       case JsNull => json.writeNull()
     }
   }
+
+  private def containsEOrDot(s: String): Boolean =
```
cchantep, line 133: There is no need for such a function

cchantep:
Unit test required

pjfanning:
> Unit test required

@cchantep I think the existing tests do a good job of testing this code change.

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -477,6 +477,19 @@ class JsonSpec extends org.specs2.mutable.Specification {
       fromJson[JsonNode](jsNum).map(_.toString).must_==(JsSuccess("12.345"))
     }
 
+    "Serialize JsNumbers with e notation correctly" in {
+      val numStrings = Seq(
+        "1.23456789012345679012345679e999",
+        "-1.23456789012345679012345679e999",
+        "1.23456789012345679012345679e-999",
+        "-1.23456789012345679012345679e-999"
```
cchantep, line 485: Not only "positive" test about decimal case must be covered, but the other non decimal case

pjfanning: more test cases added - also to #1077 to show that the cases already pass without this set of changes

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
```diff
@@ -96,10 +95,15 @@ private[jackson] class JsValueSerializer(jsonConfig: JsonConfig) extends JsonSer
...
 
-        if (raw.indexOf('E') < 0 && raw.indexOf('.') < 0)
-          json.writeTree(new BigIntegerNode(new BigInteger(raw)))
+        if (raw.exists(c => c == 'E' || c == '.'))
+          json.writeNumber(raw)
         else
-          json.writeTree(new DecimalNode(new JBigDec(raw)))
+          json match {
+            case _: TokenBuffer =>
+              json.writeTree(new BigIntegerNode(new BigInteger(raw)))
```
pjfanning, line 103: I have a change in jackson 2.18 which means this line can be changed to something that avoids the BigInteger parsing.

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -477,6 +478,55 @@ class JsonSpec extends org.specs2.mutable.Specification {
...
+        "1",
+        "-1",
+        Int.MaxValue.toString,
+        Int.MinValue.toString,
+        Long.MaxValue.toString,
+        Long.MinValue.toString,
+        BigInteger.valueOf(Long.MaxValue).add(BigInteger.ONE).toString,
+        BigInteger.valueOf(Long.MinValue).add(BigInteger.valueOf(-1)).toString
+      )
+      numStrings.map { numString =>
```
cchantep, line 493:
```suggestion

      numStrings.map { numString =>
```

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -477,6 +478,55 @@ class JsonSpec extends org.specs2.mutable.Specification {
...
+        val bigDec = new java.math.BigDecimal(numString)
+        Json.stringify(JsNumber(bigDec)).must_==(bigDec.toString)
```
cchantep, line 495:
```suggestion
        Json.stringify(JsNumber(bigDec)) must_=== numStr
```

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -477,6 +478,55 @@ class JsonSpec extends org.specs2.mutable.Specification {
...
+        "1.23456789",
+        "-1.23456789",
+        Float.MaxValue.toString,
+        Float.MinValue.toString,
+        Double.MaxValue.toString,
+        Double.MinValue.toString,
+        java.math.BigDecimal.valueOf(Double.MaxValue).add(java.math.BigDecimal.valueOf(1)).toString,
+        java.math.BigDecimal.valueOf(Double.MinValue).add(java.math.BigDecimal.valueOf(-1)).toString
+      )
+      numStrings.map { numString =>
```
cchantep, line 511: same

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
```diff
@@ -477,6 +478,55 @@ class JsonSpec extends org.specs2.mutable.Specification {
...
+    }
+
+    "Serialize JsNumbers with e notation correctly" in {
+      val numStrings = Seq(
+        "1.23456789012345679012345679e999",
+        "-1.23456789012345679012345679e999",
+        "1.23456789012345679012345679e-999",
+        "-1.23456789012345679012345679e-999"
+      )
+      numStrings.map { numString =>
```
cchantep, line 524: same

play-json/jvm/src/test/scala/play/api/libs/json/JsonSpec.scala
pjfanning: fails with this when forced to check the strings are equal as opposed to the numbers are equal

```
[info]   + Serialize JsNumbers with integers correctly (14 ms)
[error]   x Serialize JsNumbers with decimal points correctly (10 ms)
[error]    '3.4028235E+38' != '3.4028235E38' (JsonSpec.scala:513)
[error]   x Serialize JsNumbers with e notation correctly (10 ms)
[error]    '1.23456789012345679012345679E+999' != '1.23456789012345679012345679e999' (JsonSpec.scala:526)
```

This also happens with #1077 which has these tests but not the main source change

pjfanning:
@mkurz I can make a small adjustment in this PR because newer jackson releases have a new method that I added so this PR can use it

play-json/jvm/src/main/scala/play/api/libs/json/jackson/JacksonJson.scala
mkurz: @pjfanning I merged the test (#1077) and also the main branch is on Jackson 2.19.1 now thanks to your pull request. Can you please now update this line with what you mentioned (as we are now on Jackson 2.18+)? thanks!

mkurz:
> @mkurz I can make a small adjustment in this PR because newer jackson releases have a new method that I added so this PR can use it

haha, I just commented in the code ;) yes please go ahead

(I guess it will be
```
            case tb: TokenBuffer =>
              tb.writeNumber(raw, true)
```
because I was looking at your fork already ;)
)

pjfanning:
> > @mkurz I can make a small adjustment in this PR because newer jackson releases have a new method that I added so this PR can use it
> 
> haha, I just commented in the code ;) yes please go ahead
> 
> (I guess it will be
> 
> ```scala
>             case tb: TokenBuffer =>
>               tb.writeNumber(raw, true)
> ```
> 
> because I was looking at your fork already ;) )

I pushed that change.

mkurz:
@pjfanning Looking at your fork, do you also want to PRs for the rest?
- optional support for Jackson's fast decimal parser
- StreamWriteConstraints

?

PR #1175, opened 2025-06-18
Add support for Scala 3 type class derivation of `Format`, `Writes` and `Reads`

### LINKED ISSUE #1089: [Feature] Support typeclass derivation (`derives`) from Scala 3
_This is a feature request, not an issue._

Following the discussion on this [Reddit thread](https://www.reddit.com/r/scala/comments/1eknq5i/derived_type_class_for_json_formatting/), it would be nice if it was supported OOB by play-json. I'm copying the relevant part of the thread here.

From @SnooDogs8860:

>  In Scala 2.13, and play-json, I used the following pattern to serialize/de-serialize between case classes and Json:
>
> ```scala
> import play.api.libs.json.{Format, Json}
> 
> object WelcomeResult {
>   implicit val format = Json.format[WelcomeResult]
> }
> 
> case class WelcomeResult(service: String)
> ```
> 
> defining implicit formatin the companion object of the case class, made it easily visible from outside, so the following code worked without > even requiring an extra import:
> 
> ```scala
> val welcome = WelcomeResult("test")
> val json = Json.toJson(welcome)
> ```
> 
> In Scala 3 I saw the type class derivation feature and though I could take advantage of it by a. making the code more compact. b. defining such a default formatting only once in a generic way in the type class.
> 
> This is the implementation I have in mind:
> 
> ```scala
> case class WelcomeResult(service: String) derives Format 
>  ```

From @mkatrenik and @SnooDogs8860, the implementation could like the following (using another custom `trait` as modifying the PlayJSON one was not possible):

```scala
trait JsonFormat[T] extends Format[T]

object JsonFormat:

  inline def derived[T]: JsonFormat[T] =
    new JsonFormat[T]:
      override def reads(json: JsValue): JsResult[T] = Json.reads[T].reads(json)

      override def writes(o: T): JsValue = Json.writes[T].writes(o)
```

---

_It might only be a matter of opening a PR with the code above, I don't know how much changes would be required in play-json code to integrate it. I guess most of the "issue" is to have it only for Scala 3 and avoiding duplication of having a source file for Scala 2 and one for Scala3?_

DavidGregory084:
# Pull Request Checklist

## Fixes

Fixes #1089.

## Purpose

This PR adds support for Scala 3's [Type Class Derivation](https://docs.scala-lang.org/scala3/reference/contextual/derivation.html) when using `Reads`, `Writes` and `Format`.

For example:

```scala
case class Resident(name: String, age: Int, role: Option[String]) derives Reads
```

This is done by providing an `inline def derived` in the companion object of `Reads`, `Writes` and `Format` that simply forwards to the appropriate `Json` macro implementation.  

## Background Context

Using an `inline def` should ensure that `derives Reads` behaves the same way as adding a `given` declaration to the class's companion object:

```scala
object Resident {
  given Reads[Resident] = Json.reads[Resident]
}
```

This expansion is described in the [documentation](https://docs.scala-lang.org/scala3/reference/contextual/derivation.html#f-and-all-arguments-of-derivingtype-have-kind-) for type class derivation, especially the example for `Point`.

## References

Please see the [Reddit discussion](https://www.reddit.com/r/scala/comments/1eknq5i/derived_type_class_for_json_formatting/) linked in issue #1089.

### DIFF
--- a/docs/manual/working/scalaGuide/main/json/code-3/Scala3JsonAutomatedSpec.scala
+++ b/docs/manual/working/scalaGuide/main/json/code-3/Scala3JsonAutomatedSpec.scala
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) from 2022 The Play Framework Contributors <https://github.com/playframework>, 2011-2021 Lightbend Inc. <https://www.lightbend.com>
+ */
+
+package scalaguide.json
+
+import org.specs2.mutable.Specification
+import play.api.libs.json._
+
+object Scala3JsonAutomatedSpec {
+  //#reads-model
+  case class Resident(
+    name: String,
+    age: Int,
+    specialism: Option[String]
+  ) derives Reads
+  //#reads-model
+
+  val residentJson = Json.parse(
+    """{
+      "name" : "Fiver",
+      "age" : 4
+    }"""
+  )
+
+  val sampleResident = Resident("Fiver", 4, None)
+
+  //#reads-trait
+  sealed trait Role derives Reads
+  case object Admin                            extends Role derives Reads
+  case class Contributor(organization: String) extends Role derives Reads
+  //#reads-trait
+
+  val contributorJson = Json.obj("_type" -> "scalaguide.json.Scala3JsonAutomatedSpec.Contributor", "organization" -> "Foo")
+  val sampleContributor = Contributor("Foo")
+}
+
+class Scala3JsonAutomatedSpec extends Specification {
+  import Scala3JsonAutomatedSpec._
+
+  "Scala 3 JSON automated" should {
+    "for case class" >> {
+      "derive a Reads" in {
+        residentJson.as[Resident].must_===(sampleResident)
+      }
+    }
+    "for trait" >> {
+      "derive a Reads if every subclass derives Reads" in {
+        contributorJson.as[Role].must_===(sampleContributor)
+      }
+    }
+  }
+}
--- a/play-json/shared/src/main/scala/play/api/libs/json/Format.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/Format.scala
@@ -64,7 +64,7 @@ object OFormat {
 /**
  * Default Json formatters.
  */
-object Format extends PathFormat with ConstraintFormat with DefaultFormat {
+object Format extends PathFormat with ConstraintFormat with DefaultFormat with DerivedFormat {
   val constraints: ConstraintFormat = this
   val path: PathFormat              = this
 
--- a/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
@@ -237,7 +237,7 @@ object OWrites extends PathWrites with ConstraintWrites {
 /**
  * Default Serializers.
  */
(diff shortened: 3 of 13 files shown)


DavidGregory084:
Hmm seeing a test failure on `JsonSpec` which seems unrelated to these changes:

> [error]       x preserve a single trailing zero decimal with preserveZeroDecimal=true (40 ms)
> [error]        '1' != '1.0' (JsonSpec.scala:352)

gaeljw:
That looks really great. Scala 3 users will be happy to get this :)

play-json/shared/src/test/scala-3/play/api/libs/json/DerivesSyntaxSpec.scala
```diff
@@ -0,0 +1,97 @@
...
+import org.scalatest.matchers.must.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+final class DerivesSyntaxSpec extends AnyWordSpec with Matchers with EitherValues {
+  "Derives syntax" should {
+    "derive Format using macros" when {
+      "used with standalone case classes" in {
+        val input = DerivesFormat.StandaloneCaseClass(1)
+        val json  = Json.toJson(input)
+        json mustBe Json.obj("intField" -> 1)
```
cchantep, line 17:
```suggestion

        json mustBe Json.obj("intField" -> 1)
```

docs/manual/working/scalaGuide/main/json/code-3/Scala3JsonAutomatedSpec.scala
```diff
@@ -0,0 +1,53 @@
...
+class Scala3JsonAutomatedSpec extends Specification {
+  import Scala3JsonAutomatedSpec._
+
+  "Scala 3 JSON automated" should {
+    "for case class" >> {
+      "derive a Reads" in {
+        residentJson.as[Resident].must_===(sampleResident)
+      }
+    }
+    "for trait" >> {
```
cchantep, line 47:
```suggestion

    "for trait" >> {
```

cchantep (commented):
Aside cosmetics comment, lgtm

play-json/shared/src/main/scala/play/api/libs/json/Format.scala
```diff
@@ -64,7 +64,7 @@ object OFormat {
 /**
  * Default Json formatters.
  */
-object Format extends PathFormat with ConstraintFormat with DefaultFormat {
+object Format extends PathFormat with ConstraintFormat with DefaultFormat with DerivedFormat {
```
cchantep: I guess derivation should be reserved for `OFormat` (see Scala 3 implementation)

play-json/shared/src/main/scala/play/api/libs/json/Writes.scala
```diff
@@ -237,7 +237,7 @@ object OWrites extends PathWrites with ConstraintWrites {
 /**
  * Default Serializers.
  */
-object Writes extends PathWrites with ConstraintWrites with DefaultWrites with GeneratedWrites {
+object Writes extends PathWrites with ConstraintWrites with DefaultWrites with GeneratedWrites with DerivedWrites {
```
cchantep: I guess derivation should be reserved for `OWrites` (see Scala 3 implementation)

Issue #1019, opened 2024-04-29
Json.writes[myScala3Enum] crashes play-json / the compilation?


nemoo:
Since somebody else had the same problem, I am now opening this ticket. I guess lots of users who are now starting to use scala 3 will run into this problem.

**Expected behaviour**
If play-json currently does not support scala 3 enums via Json.writes, it should tell the user that in a compiler message.

**Actual behaviour**
If Json.writes[T] is used on a scala 3 enum, compiling it with scala-cli or sbt crashes the compilation process. Compilation can only be killed via the task manager.

How to reproduce (tested on Java 11):
```scala
//> using scala 3.3.3
//> using dep org.playframework::play-json:3.0.2
import play.api.libs.json.*

enum Color:
  case red,green,blue

@main
def main =
  implicit val myWrites: OWrites[Color] = Json.writes[Color] //this line produces the crash
```

cchantep:
Duplicate of #1017

PR #1176, opened 2025-06-25
For information / discussion: Add support for Scala 3 enums

### LINKED ISSUE #1017: Support Scala 3 Enums
Let's be honest, this is a feature request, not a bug report. 

Play-json currently supports Enumerations out of the box via methods like `Json.formatEnum(X)`
https://github.com/playframework/play-json/blob/main/play-json/shared/src/main/scala/play/api/libs/json/Reads.scala#L437

It would be great if the current Scala 3 way of doing enums: https://docs.scala-lang.org/scala3/reference/enums/enums.html would also be supported in a similar fashion.

DavidGregory084:
# Pull Request Checklist

## Fixes

Fixes #1017.

However, it does so using Scala 3's [Mirror.Sum](https://docs.scala-lang.org/scala3/reference/contextual/derivation.html). This forbids any subclasses which are not products or sums themselves.

This is why the test for `knownSubclasses` is failing - `TestUnion.UT` has a subclass `UC` which is a normal class.

## Purpose

This PR adds support for Scala 3 `enum`s.

## Background Context

The problem with the current implementation is that it assumes that the `.children` of a `TypeRepr` will be subclasses of that `TypeRepr`.

Scala 3 `enum` `case`s are encoded as `val`s which are bound to an anonymous class. They are ascribed with the type of the parent definition.

For example:

```scala
enum Color {
  case Red, Green, Blue
}

// Color.Red is encoded as ===>

val Red: Color = $new(0, "Red")
```

This causes the `subclasses` function of the `knownSubclasses` helper to loop indefinitely trying to find the subclasses of `Color` because the `tpt.tpe` of the `ValDef` children is the same `TypeRepr` for `Color` that we started with. This is the cause for #1019.

I do not know how to get the singleton `TypeRepr` of the `enum` `case` from the `ValDef`, but I think that would be the way to solve this problem while preserving the existing behaviour of the macro.

## References

[Scala 3 Enumerations Implementation](https://docs.scala-lang.org/scala3/reference/enums/enums.html#implementation)

### DIFF
--- a/play-json/shared/src/main/scala-3/play/api/libs/json/JsMacroImpl.scala
+++ b/play-json/shared/src/main/scala-3/play/api/libs/json/JsMacroImpl.scala
@@ -241,8 +241,12 @@ object JsMacroImpl { // TODO: debug
                   TypeRepr.of[String]
                 )
 
+                val termSym = tpr.termSymbol
+                val typeSym = tpr.typeSymbol
+                val isCase = termSym.flags.is(Flags.Case)
+
                 val tpeCaseName: Expr[String] = '{
-                  ${ config }.typeNaming(${ Expr(typeName(tpr.typeSymbol)) })
+                  ${ config }.typeNaming(${ Expr(typeName(if isCase then termSym else typeSym)) })
                 }
 
                 val resolve = resolver[Reads, sub](
@@ -595,8 +599,12 @@ object JsMacroImpl { // TODO: debug
                   subTpr
                 )
 
+                val termSym = tpr.termSymbol
+                val typeSym = tpr.typeSymbol
+                val isCase = termSym.flags.is(Flags.Case)
+
                 val tpeCaseName: Expr[String] = '{
-                  ${ config }.typeNaming(${ Expr(typeName(tpr.typeSymbol)) })
+                  ${ config }.typeNaming(${ Expr(typeName(if isCase then termSym else typeSym)) })
                 }
 
                 val resolve = resolver[Writes, sub](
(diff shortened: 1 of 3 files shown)


play-json/shared/src/main/scala-3/play/api/libs/json/QuotesHelper.scala
```diff
@@ -43,51 +42,27 @@ private[json] trait QuotesHelper {
...
-        }
+        case '[EmptyTuple] => Nil
       }
 
-      val types = subclasses(cls.children.map(_.tree), Nil)
-
-      if types.isEmpty then None else Some(types)
+    tpr.asType match {
+      case '[t] =>
+        Expr.summon[Mirror.Of[t]].collect {
```
DavidGregory084, line 60: I think that a less WIP implementation of this should probably use `Implicits.search` so that it could surface any implicit search error messages to the user

play-json/shared/src/main/scala-3/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -241,8 +241,12 @@ object JsMacroImpl { // TODO: debug
                   TypeRepr.of[String]
                 )
 
+                val termSym = tpr.termSymbol
+                val typeSym = tpr.typeSymbol
+                val isCase = termSym.flags.is(Flags.Case)
+
                 val tpeCaseName: Expr[String] = '{
-                  ${ config }.typeNaming(${ Expr(typeName(tpr.typeSymbol)) })
+                  ${ config }.typeNaming(${ Expr(typeName(if isCase then termSym else typeSym)) })
```
DavidGregory084, line 249: By using the `termSymbol` we get the fully qualified name of the `enum` `case` rather than the name of the `enum` as our type discriminator.

PR #1204, opened 2025-07-10
Make JacksonJson `private[play]` so we can use it in Play

mkurz:
/cc @pjfanning 

I don't think there is the need to do the same for the companion object (?)


mkurz:
Releasing as `M3` so we can make use of it in Play: https://github.com/playframework/play-json/actions/runs/16195394512/job/45720500379

PR #1268, opened 2025-12-14
🧹 chore: upgrade to Jackson 3

### LINKED ISSUE #1258: Upgrade to Jackson 3
As per the title.

There's definitely no urgency in doing it but it'll be good to help the ecosystem move forward.

Jackson 2.x and 3.x can co-exist, so I don't think there's a reason for play-json to wait. Users will still be able to use other libraries relying on Jackson 2.x.

Maybe this is something we can include in the next minor release?

[Migration notes](https://github.com/FasterXML/jackson/blob/main/jackson3/MIGRATING_TO_JACKSON_3.md)

_I'll be happy to work on a PR for this._

gaeljw:
# Pull Request Checklist

## Fixes

Fixes #1258

## Purpose

Upgrade Jackson from 2.x to 3.x.

## Background Context

- `build.sbt`: I merged declaration of `jackson-databind` with the other Jackson dependencies, not sure why it was separate but there doesn't seem to be any reason anymore
- `JacksonJson`:
  - I had to change `ObjectMapper` to `JsonMapper` so that we can use `.enable(JsonWriteFeature.ESCAPE_NON_ASCII)` (not available on `ObjectMapper` as it's a JSON specific feature). This makes sense anyway IMHO _but_ it'll be a source breaking change for users of `setObjectMapper` (Play Framework?)
  - `generateFromJsValue` rebuild a new `JsonMapper` at each call, this is very likely not optimal. We could store an instance of mapper with the proper escapeNonAscii configuration in addition to the already stored mapper.
  - removed `createGenerator`, they don't bring value since we're writing a single value and they duplicated some configuration (escapeNonAscii, pretty printer)
  - `serialize` method for `JsNumber`: for some reason, Jackson `JsonGenerator.writeNumber(String)` now writes the number as a string (with extra quotes), at least it what's the [`TreeBuildingGenerator` implementation actually used does](https://github.com/FasterXML/jackson-databind/blob/f946c3d8db48fabe514a912d9dec856d0f659f97/src/main/java/tools/jackson/databind/node/TreeBuildingGenerator.java#L408), this was visible in `JsonSpec` tests. In order to workaround that, I had to use `writeNumber(BigDecimal)` instead but this means an extra `BigDecimal` allocation.
    - When looking at other implementations of `JsonGenerator.writeNumber(String)`, I wonder if this the `TreeBuildingGenerator` implementation is really expected: others do write as a "raw" number.
    - ~~I couldn't find a change in 3.x that explains this behaviour change though. The implementations in Jackson are like this for quite some time already.~~ EDIT: this is actually caused by the removal of `createGenerator` steps, it was previously using the [`WriterBasedJsonGenerator` implementation](https://github.com/FasterXML/jackson-core/blob/0406a51a186125a451e3353ee1824f9a0407b152/src/main/java/tools/jackson/core/json/WriterBasedJsonGenerator.java#L921)
- `JsonSpec`: added assertions on the resulting type when deserializing from `JsValue` (Play JSON) to `JsonNode` (Jackson) to better cover the intent of the test cases. Asserting only on the string representation is quite risky (I actually had a working implementation where the string representation was good but the Jackson nodes were not making sense).
- 3 modules are now included by default in Jackson: `ParameterNamesModule`, `Jdk8Module`, `JavaTimeModule`, hence we no longer need to enable them explicitly (we couldn't anyway)

## Remaining work

- [ ] Decide if it's best to optimize `generateFromJsValue` to not rebuild a new `JsonMapper` each time
- [ ] Refine/rediscuss the changes around `JsNumber` serialization and `JsonGenerator.writeNumber(String)`
- [ ] Open issue and PR in play-framework to adapt the usage of `setObjectMapper`

## References

- https://github.com/FasterXML/jackson/blob/main/jackson3/MIGRATING_TO_JACKSON_3.md

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,19 +21,12 @@ def specs2(scalaVersion: String) =
     ("org.specs2" %% s"specs2-$n" % "4.23.0") % Test
   }
 
-val jacksonDatabindVersion = "2.20.1"
-val jacksonDatabind        = Seq(
-  "com.fasterxml.jackson.core" % "jackson-databind" % jacksonDatabindVersion
-)
-
-val jacksonVersion = jacksonDatabindVersion
+val jacksonVersion = "3.0.3"
 val jacksons       = Seq(
-  "com.fasterxml.jackson.core"     % "jackson-core",
-  "com.fasterxml.jackson.datatype" % "jackson-datatype-jdk8",
-  "com.fasterxml.jackson.datatype" % "jackson-datatype-jsr310",
-  "com.fasterxml.jackson.module"   % "jackson-module-parameter-names",
-  "com.fasterxml.jackson.module"  %% "jackson-module-scala",
-).map(_ % jacksonVersion) ++ jacksonDatabind
+  "tools.jackson.core"    % "jackson-core",
+  "tools.jackson.core"    % "jackson-databind",
+  "tools.jackson.module" %% "jackson-module-scala",
+).map(_ % jacksonVersion)
 
 val joda = Seq(
   "joda-time" % "joda-time" % "2.14.0"
--- a/play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/EnvReads.scala
@@ -26,9 +26,9 @@ import java.util.Locale
 
 import scala.util.control.NonFatal
 
-import com.fasterxml.jackson.databind.JsonNode
-import com.fasterxml.jackson.databind.node.ArrayNode
-import com.fasterxml.jackson.databind.node.ObjectNode
+import tools.jackson.databind.JsonNode
+import tools.jackson.databind.node.ArrayNode
+import tools.jackson.databind.node.ObjectNode
 
 import play.api.libs.json.jackson.JacksonJson
 
--- a/play-json/jvm/src/main/scala/play/api/libs/json/EnvWrites.scala
+++ b/play-json/jvm/src/main/scala/play/api/libs/json/EnvWrites.scala
@@ -18,7 +18,7 @@ import java.time.ZonedDateTime
 import java.time.{ Duration => JDuration }
 import java.util.Locale
 
-import com.fasterxml.jackson.databind.JsonNode
+import tools.jackson.databind.JsonNode
 import play.api.libs.json.jackson.JacksonJson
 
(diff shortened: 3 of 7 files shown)


gaeljw:
:warning: For now, 4 tests are failing for deserializing `JsNumber` to Jackson number node because they currently deserialize to a Jackson `StringNode`

pjfanning:
@gaeljw @mkurz Jackson 2 is stable and still supported. I don't see much reason to move to Jackson 3 yet. It could take a while for it to become stable. In Pekko, I'm retaining Jackson 2 support while adding Jackson 3 support in Pekko 2 as an option.
I don't know what the intent is here but I think dropping Jackson 2 and moving to Jackson 3 has its risks and difficulties.

gaeljw:
The intent is only to be "up to date". I'm a bit surprised when you say Jackson 3 might not be stable yet: it's the default in Spring Boot for a few weeks now, a couple of libraries I know have already switched to Jackson 3... Do you have something specific in mind or it's just the fact that it's a major version and takes time to get feedbacks and detect issues?

I don't think there's any major reason to upgrade as soon as possible but on the other hand I don't see a reason to wait. The next minor version of Play could be a good candidate for the upgrade, isn't it?

Especially, as Jackson 2 and 3 can live together in the same codebase, this doesn't force users to have all their dependencies aligned on Jackson 3.

FWIW, we've been using Jackson 3 + Jackson 2 (both present at runtime, Play using 2.x and other libraries using 3.x) at work in a few projects for a few weeks and didn't have any issue so far.

pjfanning:
@mkurz @gaeljw Jackson 3 is now onto v3.2.1 and can be regarded as being as stable as Jackson 2 latest releases.
Pekko 2 has milestone releases that have pekko-serialization-jackson3 libs that provide Jackson3 based serde support - and it should be easy enough for playframework to use that and switch to Jackson 3 too.
Akka may not have anything for Jackson 3 though. So switching to Jackson 3 may cause issues for playframework versions that are based on Akka libs.
Jackson 2 is maintained so switching to Jackson 3 is not absolutely required but there is an argument to try to switch to latest Jackson release.

gaeljw:
I can spend a bit of time to rebase the PR on top of the latest changes (both Play JSON & Jackson) so that the status is up to date, I believe a few things changed in the meantime.

PR #1413, opened 2026-08-31
Close #955 - Map key constraint

### LINKED ISSUE #955: `LowPriorityKeyReads.readableKeyReads` should not exist

3.0.1

Scala

### Problem

Map serialization/deserialization is broken if the key doesn't have a `KeyReads` instance.

The problem stems from the `LowPriorityKeyReads.readableKeyReads` which implies that if you have a `Reads[A]` then you have `KeyReads[A]`, which does not seem like a logical conclusion. After all, the `Reads` can expect `JsArray`, `JsObject` or any other type, none of which are `JsString`.

```scala
import play.api.libs.json.*

val map = Map((1, 2) -> "foo")

val asJson = Json.prettyPrint(Json.toJson(map)) 
println(asJson)
// Gets serialized to: [ [ [ 1, 2 ], "foo" ] ]

val fromJson = Json.parse(asJson).validate[Map[(Int, Int), String]]
println(fromJson)
// Fails to deserialize, as a failing `KeyReads[(Int, Int)]` will be created from `LowPriorityKeyReads.readableKeyReads`
```

### Suggested solution

Remove `LowPriorityKeyReads.readableKeyReads`. This breaks backwards compatibility, but I argue that this function is useless (as in it almost never produces a useful `Reads` instance anyway).

### Reproducible Test Case

https://scastie.scala-lang.org/8ePgBH6lQ5i4GmCOPksvyw

cchantep:
# Pull Request Checklist

## Fixes

Fixes #955

## Purpose

Fix implicit resolution of `KeyReads` / `KeyWrites` for types represented as JSON strings.

`readableKeyReads` currently derives a `KeyReads[T]` from any `Reads[T]`. This is too broad: types such as tuples have a `Reads`, but are not represented as JSON strings and therefore cannot be used as JSON object keys through this mechanism.

This PR introduces `Format.Representation[T, R]` as explicit evidence that a type has a given JSON representation, and uses it to constrain the string-based `KeyReads` / `KeyWrites` derivation.

The existing `readableKeyReads` is kept for backward compatibility but deprecated.

## Background Context

A `Reads[T]` does not imply that `T` can be read from a `JsString`. For example, `Reads[(Int, Int)]` exists even though `(Int, Int)` is not a string representation (unlike for example `URI` which is represented as JSON string).

The new `Representation[T, JsString]` makes this distinction explicit and prevents `KeyReads` / `KeyWrites` from being derived for types that are not string-represented.

Representations for types whose JSON representation depends on the `Writes` / `Reads` (e.g. temporal types, duration, ...) in scope are provided separately through `Format.Representation.Implicits`, so they are opt-in rather than unconditional.

### DIFF
--- a/play-json/shared/src/main/scala/play/api/libs/json/KeyReads.scala
+++ b/play-json/shared/src/main/scala/play/api/libs/json/KeyReads.scala
@@ -69,7 +69,29 @@ object KeyReads extends EnvKeyReads with LowPriorityKeyReads {
 }
 
 private[json] sealed trait LowPriorityKeyReads {
-  implicit def readableKeyReads[T](implicit r: Reads[T]): KeyReads[T] =
+
+  /**
+   * Creates a [[KeyReads]] for values represented as JSON strings.
+   *
+   * The supplied [[Reads]] is applied to a [[JsString]] containing the object key.
+   * A [[Format.Representation]] is required to explicitly indicate that `T` is
+   * represented as a JSON string.
+   *
+   * @tparam T the type to read
+   * @return a [[KeyReads]] for `T`
+   */
+  implicit def stringRepresentedKeyReads[T](implicit
+      r: Reads[T],
+      repr: Format.Representation[T, JsString]
+  ): KeyReads[T] =
+    KeyReads[T] { key =>
+      r.reads(JsString(key))
+    }
+
+  @deprecated(
+    "Use stringRepresentedKeyReads instead. A Format.Representation[T, JsString] is required to ensure that T is represented as a JSON string.",
+    "3.1"
+  ) def readableKeyReads[T](implicit r: Reads[T]): KeyReads[T] =
     KeyReads[T] { key =>
       r.reads(JsString(key))
     }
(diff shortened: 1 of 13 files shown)


play-json/shared/src/main/scala/play/api/libs/json/KeyReads.scala
```diff
@@ -69,7 +69,29 @@ object KeyReads extends EnvKeyReads with LowPriorityKeyReads {
...
+  implicit def stringRepresentedKeyReads[T](implicit
+      r: Reads[T],
+      repr: Format.Representation[T, JsString]
+  ): KeyReads[T] =
+    KeyReads[T] { key =>
+      r.reads(JsString(key))
+    }
+
+  @deprecated(
+    "Use stringRepresentedKeyReads instead. A Format.Representation[T, JsString] is required to ensure that T is represented as a JSON string.",
```
cchantep, line 92: For compatibility

play-json/shared/src/test/scala/play/api/libs/json/FormatSharedSpec.scala
```diff
@@ -27,9 +28,57 @@ final class FormatSharedSpec extends AnyWordSpec with Matchers {
       val expectedRepr = Json.obj("name" -> "bar")
 
       strFormat.reads(expectedRepr).mustEqual(JsSuccess("bar"))
+
       strFormat.writes("bar").mustEqual(expectedRepr)
     }
   }
 
+  "Representation" should {
+    "not be found by default" in {
```
cchantep, line 37: Temporal types *can* be represented as string, but do not force it by default.

Issue #367, opened 2019-12-07
Annotations for macro similar to @Flatten & @Ignore


cchantep:
http://reactivemongo.org/releases/0.1x/documentation/bson/typeclasses.html

ktham:
@cchantep I see that [#1427](https://github.com/playframework/play-json/pull/1427) implements this, nice!

Would you also consider an annotation that redacts fields, something like @Redact("<redacted>")? This would help where the key needs to stay in the JSON object to preserve the structure, but where value shouldn't be serialized.

For example, I'd like `case class UserCredentials(username: String, @Redact("*****") password: String)` to serialize `UserCredentials("user1", "pA$$word")` into:

```json
{
  "username" : "user1",
  "password" : "*****"
}
```

Right now there are places where I'm hand-writing a play-json Writes to do this, but it would be nice to be able to customize the macro with an annotation, similar to what [#1427](https://github.com/playframework/play-json/pull/1427) is doing.

Happy to open a separate issue for this

PR #1427, opened 2026-09-18
Close #367 - Introduce annotations @ignore and @flatten for macro usages

cchantep:
# Pull Request Checklist

## Purpose

Introduce `@Ignore` and `@Flatten` field annotations for automated JSON mapping (Scala 2/3 macros), addressing #367.

Macros (`Json.reads` / `Json.writes` / `Json.format`, and Scala 3 `derives`) can customize case-class fields without hand-written `Format`s:

- **`Json.Annotations.Ignore`** (also `@transient`): omit the field on write; on read use a constructor default or `None` for bare `Option` (compile error otherwise). Defaults apply for `@Ignore` even without `Json.DefaultValues`.
- **`Json.Annotations.Flatten`**: merge a nested **JSON object** into the parent instead of nesting under the property name. Nested writers must be `OWrites`/`OFormat` (checked at compile time). Recursive self-types are rejected.
- **`@Flatten Option[T]`**: write `Some` merges / `None` omits; read maps success to `Some` and **propagates** nested `JsError`s (same as non-`Option` flatten).

Covered by `MacroSpec` and the automated JSON guide (`ScalaJsonAutomated.md` + examples).

## Background Context

Approach follows ReactiveMongo-BSON-style field annotations (`Ignore` / document flatten), scoped as an **MVP** for play-json:

- Public API under `Json.Annotations`, wired into existing Scala 2 and Scala 3 macro backends.
- Flatten needs a real nested object → require `OWrites` at expand time (no runtime `JsObject` match / `IllegalArgumentException`).
- `@Flatten Option` read defaults to **error propagation** (`.map(Some(_))`) so invalid nested data is not silently turned into `None`.
- Sealed-family forward codecs stay typed as the parent `M[A]` (Placeholder self-refs), avoiding unsound `asInstanceOf`/`asExprOf[M[sub]]`.
- Scala 2.12-safe collections (e.g. `.zipped` instead of `lazyZip`).

**Out of scope for this MVP:** `@Key` / `@DefaultValue`, compile-time key-collision detection, full BSON parity (documented in scaladoc and the guide).

## References

- Closes **#367**
- Design inspiration: ReactiveMongo-BSON `MacroAnnotations` / `MacroImpl` (`@Ignore`, `@Flatten` document merge)


play-json/shared/src/main/scala-2/play/api/libs/json/JsMacroImpl.scala
```diff
@@ -679,34 +680,184 @@ class JsMacroImpl(val c: blackbox.Context) {
...
+            ann.tree.tpe =:= typeOf[scala.transient]
+        )
+
+      def hasFlatten(param: Symbol): Boolean =
+        param.annotations.exists(_.tree.tpe =:= typeOf[Json.Annotations.Flatten])
+
+      /**
+       * `@Flatten` write path needs a nested object. Prefer the already-resolved
+       * `impl` when it is already an `OWrites`/`OFormat`; otherwise require an
+       * `OWrites` in scope (compile-time, like BSONDocumentWriter checks).
```
ktham: Should we remove the reference to `BSONDocumentWriter`?

ktham, line 695: Though it was inspired from it, should we remove the reference to BSONDocumentWriter?

(I'm really looking forward to seeing the ability to customize the macros in the next play-json release!)

